Issue #3, opened 2021-06-25
Create frontend codebase scaffolding


longtomjr:
- [x] Create Angular project - #7
 - [x] #9
 - [x] Use mobile 1st component library 
 - [x] Setup Storybook - #8
 - [x] Add empty pages for the views - #15 

> Empty views to setup:
> Landing page:
>  - Choose action: Create Wallet, Open Wallet
> 
> Create Wallet:
>  - Take creation details
>  - Show result and/or redirect to Opened Wallet
> 
> Open Wallet:
>  - Take wallet details
>  - Show errors, or redirect to Opened Wallet
> 
> Opened Wallet:
>  - Show wallet details (name, balance)
>  - Actions: Show QR code (for receiving), Send Funds
> 
> Send Funds:
>  - Take recipient details (from text field or scanning QR code)
>  - Show success / failure
>  - Refresh balance (?)
>  - Next: Opened Wallet

PiDelport:
This should be done with #15. 🎉

PR #8, opened 2021-06-29
add Storybook

PiDelport:
Issues:

- #3
- #10

Depends on:

- #9


PiDelport:
Hmm, it looks like one of the Storybook dependencies in 9e7dc54 breaks `build:prod` (but not `build`).

PiDelport:
### Investigation

It looks like this is one _those_ npm issues… 🤦🏼‍♀️

At first, I had trouble reproducing this reliably: a normal `npm install` would not trigger the crash after updating to the breaking changeset (or stop the crash, after switching to the parent changeset), but a full `npm ci` would.

In theory, the result of those two commands should be the same, but they're not. 

To inspect closer, 🕵🏼‍♀️ I compared snapshots of the working and non-working `node-modules`, and apparently the breaking difference is that the contents of the `cssnano-utils` package somehow gets replaced with the contents of the `postcss-values-parser` package (which doesn't seem related), in the same directory. 🤯

It looks like something went wrong in `package-lock.json`:

```diff
     "cssnano-utils": {
       "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/cssnano-utils/-/cssnano-utils-2.0.1.tgz",
-      "integrity": "sha512-i8vLRZTnEH9ubIyfdZCAdIdgnHAUeQeByEeQ2I7oTilvP9oHO6RScpeq3GsFUVqeB8uZgOQ9pw8utofNn32hhQ==",
+      "resolved": "https://registry.npmjs.org/postcss-values-parser/-/postcss-values-parser-2.0.1.tgz",
+      "integrity": "sha512-2tLuBsA6P4rYTNKCXYG/71C7j1pU6pK503suYOmn4xYrQIzW+opD+7FAFNuGSdZC/3Qfy334QbeMu7MEb8gOxg==",
       "dev": true,
-      "requires": {}
+      "requires": {
+        "flatten": "^1.0.2",
+        "indexes-of": "^1.0.1",
+        "uniq": "^1.0.1"
+      }
     },
```

This could have happened due to some npm resolution weirdness, or a particularly unlucky automatic mis-merge alignment, or something else?

I'll try to fix this by replaying the dependency additions in a separate commit.

PiDelport:
That fixed it! 🎉

Diff: [3760107..bb95236](https://github.com/registreerocks/nautilus-wallet/compare/3760107..bb95236)

jongbonga (approved):

PR #9, opened 2021-06-29
build(web-client): add prettier

PiDelport:
- #3


PiDelport:
@jongbonga: Does this work for your setup?

jongbonga (commented):
Prettier literally changed the whole app init linting like "Angular CLI doesn't know what they were doing by initiating the project that way, let's fix it our way which is the right way"

They are so many things in this config that are not sitting well with me...

web-client/tsconfig.spec.json
```diff
@@ -3,16 +3,8 @@
...
-  "files": [
-    "src/test.ts",
-    "src/polyfills.ts"
-  ],
-  "include": [
-    "src/**/*.spec.ts",
-    "src/**/*.d.ts"
-  ]
+  "files": ["src/test.ts", "src/polyfills.ts"],
+  "include": ["src/**/*.spec.ts", "src/**/*.d.ts"]
```
jongbonga, line 9: How can this be "more readable" than its previous state? 
And will still break after a certain amount of characters or items anyway

PiDelport:
Update from our standup discussion this morning: like @longtomjr, I don't really have any strong opinion about our formatting as long as we have something that lets us mechanise it, and not have to worry about manual formatting.

I'd lean towards using the `prettier` defaults just so we don't have to worry about maintaining a custom config, but I'm also happy with a customised config (or even another formatter, as long as it gives us the same convenience).

jongbonga (approved):

longtomjr:
Just double checking @PiDelport. Is there anything holding this up still?

PiDelport:
@longtomjr: Nope, all good!

Issue #10, opened 2021-07-01
Set up static sites for dev


PiDelport:
We'll want a Netlify deployment for `web-client`.

We might want to evaluate https://www.chromatic.com/ for Storybook, rather than Netlify?

PRs:

- [x] Netlify for Angular build: #11
- [x] Chromatic for Storybook: #8

PiDelport:
These should be done, pending review.

PR #11, opened 2021-07-01
chore(web-client): add Netlify config

PiDelport:
- #10

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,0 +1,20 @@
+# Docs: https://docs.netlify.com/configure-builds/file-based-configuration/
+
+[build]
+base = "web-client"
+publish = "www"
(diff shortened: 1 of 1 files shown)


Nghondzweni:
I'm not yet familiar with netlify, but I'm having a look at it. 😅

PiDelport:
@Nghondzweni: Thanks!

The Netlify docs linked in the code should hopefully be useful to understand what's happening.

From a review perspective, the main things to check are that the right files are being deployed with the right commands, and that the fallback for Angular actually works. (That is, that visiting an Angular-routed path under the domain doesn't give a Netlify 404, but serves the base `index.html`.)

jongbonga (approved):

PR #12, opened 2021-07-05
Add wallet/account creation

Nghondzweni:


Nghondzweni:
@longtomjr @PiDelport So far I have pulled out the functionality to generate an account using a seed. The other method is generate account from_mnemonic which depends on the algonaut_mnemonic crate. Should I go ahead and pull that in too or should I try to work on the offline transactions?

longtomjr:
> @longtomjr @PiDelport So far I have pulled out the functionality to generate an account using a seed. The other method is generate account from_mnemonic which depends on the algonaut_mnemonic crate. Should I go ahead and pull that in too or should I try to work on the offline transactions?

For now just implement the `from_seed` one. We don't need anything else for account creation at the moment.

In general we just want the most bare-bones functionality that will work. So if we have 1 account creation mechanism, that is 100%.

PiDelport:
Yeah, `from_mnemonic` is mainly a mechanism to for paper wallets, which isn't really relevant to our use case. The only reason we'd want that is if we want to add support for users importing existing wallets into an enclave via mnemonic keys.

web-server/.gitignore
```diff
@@ -0,0 +1,4 @@
+### Rust ###
+# Generated by Cargo
+# will have compiled files and executables
+target
```
longtomjr, line 4: I am not 100% on the behavior of .gitignore files in sub-directories. Will this correctly ignore `web-server/target`?

web-server/algorand_core/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+const HASH_LEN: usize = 32;
+
```
longtomjr, line 2: I am not 100% on this yet, but it might make sense to just put all the algo stuff into one crate, since doing it in separate crates might add some overhead. 

The downside is that it will be less clear where the functionality that we ported over came from on the algonaught side.

I think both options here are fine, but it is probably worth thinking about the structure here.

web-server/algorand_core/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+const HASH_LEN: usize = 32;
```
longtomjr, line 1: For these crates, please add `#![no_std]` to the top of the `lib.rs` files. This tells rust to build this without attempting to link to `std` and if it fails it will give a useful error.

web-server/algorand_transaction/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+authors = ["Tsundzukani <trnghondzweni@gmail.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+algorand_core = {path = "../algorand_core", version = "0.1.0"}
+ring = "0.16.19"
+rand = "0.8.3"
```
longtomjr, line 12: We probably don't want to pull in these 2 dependencies (since both depends on std iirc), and rather parameterize the functions that use the functionality from these crates. 

There is a library that just adds that traits implemented by rand that might be no_std compatible. Maybe use that instead.

For ring: I know that there have been work on making a subset of ring no_std compatible, so it might be worth checking for that. Otherwise we should look at rather passing something (like a ring fn) from the caller.

web-server/algorand_transaction/src/lib.rs
```diff
@@ -0,0 +1 @@
+pub mod account;
```
longtomjr, line 1: Same as the other case, we want to add `#![no_std]` to the top if the lib file here to build without linking to `std`.

longtomjr (changes requested):

web-server/algorand_core/src/lib.rs
longtomjr: Having had a look at the follow up PR for this one, I think that keeping the separate crates makes sense, since there is a lot of functionality moved over from the respective crates. We can look at restructuring the directories in a follow up.

PiDelport:
We moved this porting to: https://github.com/registreerocks/algonaut-sgx

PR #16, opened 2021-07-07
added payment transaction create and sign

Nghondzweni:
- Follows #12


longtomjr (changes requested):
From a quick look this still needs a bit of work for this to either build with plain `no_std` or using the `sgx_std` crate. I would prefer `no_std` if possible.

PiDelport:
We moved this porting to: https://github.com/registreerocks/algonaut-sgx

PR #15, opened 2021-07-06
Routes

jongbonga:
Create empty pages/views

PR #14, opened 2021-07-06
Display QR-Code

jongbonga:
- #5

PR #17, opened 2021-07-14
build(web-client): npm updates, legacy-peer-deps

PiDelport:
- Follows #15 and #14

I think this finally gets our npm packages into a more stable state: it looks like the Angular ecosystem isn't really ready for the stricter NPM 7 peer-dependency handling yet, and needs legacy-peer-deps for the time being.


longtomjr (approved):

PiDelport:
@jongbonga: Let me know if this looks and works good to you, too?

jongbonga:
@PiDelport please remind me what exactly breaks > `12.0.4`?

PiDelport:
@jongbonga: That was the breakage we had with `build:prod` and `build-storybook` in #8, and partially fixed in ac7bc77: https://github.com/storybookjs/storybook/issues/15227#issuecomment-866576291 narrowed the issue down to this changed,  released in @angular-devkit/build-angular 12.0.4:

- https://github.com/angular/angular-cli/pull/21045

However, the fix in ac7bc77 required some manual package lock wrangling to avoid Angular subcomponents at different versions pulling in more than one webpack version, which triggers the problems described here:

- https://github.com/angular/angular-cli/issues/20773

(This happens because of Angular subcomponents using hardcoded versions for things like webpack, rather than ranges: in this case, some Angular 12.1.x component(s) pulled in webpack 5.39.1 alongside 5.38.1, which broke the build.)

To fix this more reliably, all of our Angular components must be from the same (or close enough to the same) Angular release, since those form a consistent set that Angular tests and maintains together.

(The reason for using `<12.0.4` rather than just `12.0.3` is that not all Angular subcomponents necessarily update their patch versions for every Angular patch release: for example, a bunch of the `@angular-eslint` packages are still at `12.0.0`.)

jongbonga (approved):

PR #20, opened 2021-07-15
refactor(ui): responsive container

jongbonga:


web-client/src/app/views/landing/landing.page.html
```diff
@@ -1,5 +1,5 @@
-<ion-content>
-  <ion-grid class="h-full">
+<ion-content color="prussian-blue">
+  <ion-grid class="h-full sm:max-w-max">
```
PiDelport, line 2: We should probably apply this globally, rather than per page?

Something like this, in `global.scss`?

```scss
.ion-page {
  // Limit default page width to Ionic's "sm" breakpoint.
  // Docs: https://ionicframework.com/docs/layout/css-utilities#ionic-breakpoints
  max-width: 576px;
  margin: auto;
}
```

jongbonga: That will restrict the whole app to a mobile viewport. 
We should give the user a responsive experience, mobile-first not mobile-only

web-client/src/app/views/landing/landing.page.html
```diff
@@ -1,5 +1,5 @@
-<ion-content>
-  <ion-grid class="h-full">
+<ion-content color="prussian-blue">
```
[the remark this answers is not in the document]
jongbonga: Do you mean `--ion-background-color` ? 
I thought about it and had it before the PR. 
I will probably apply it once I get feedback from a request I made to Abby

web-client/src/app/views/landing/landing.page.html
PiDelport: Well, currently all our views and designs are centered around a mobile form factor: it probably won't make much sense to mix full-width and limited-width views in the same app?

PiDelport: (This shouldn't prevent a more responsive experience: setting the global default keeps it mobile-first, but still allows individual views to override or extend the default for larger viewports.)

jongbonga: Had a chat with @longtomjr about it this morning.
The necessary changes will be applied.

Issue #1, opened 2021-06-25
Pull out sealing exchange functionality from the trusted compute codebase into the wallet codebase


longtomjr:
PRs:

- [x] #48
- [x] #67 

Follow-up issues:

- #88

## Tasks

 - [x] Remote attestation:
   - App gets and verifies attestation with the Enclave
   - Obtains enclave's public key

 - [x] Sealed enclave communication APIs:

   - Sealed request ("half-sealed"): 
     - Request message sealed to enclave's attested public key
     - Response message is not sealed

   - Sealed exchange ("fully-sealed"):
     - Client generates ephemeral keypair for exchange
     - Request message sealed to enclave's attested public key
     - Request sent along with client's ephemeral public key
     - Response message sealed to client's ephemeral public

 - Implementation notes:

   - On app side, these should be common JS helper functions

   - On server side, each of the above should probably correspond to
     a separate REST API endpoint and ECALL.

   - The message formats should probably all be JSON, for ease of implementation

   - For the fully-sealed exchange responses, we'll need variable-sized ECALL
     results. Since these should be mostly small, we can probably use the
     over-allocated out buffer approach? (Otherwise, OCALL or another strategy?)

PiDelport:
@longtomjr: Should we keep DCAP attestation as part of this, or split it to a follow-up task?

longtomjr:
@PiDelport I think splitting out a new issue would be better.

PiDelport:
Cool, I split out:

- #88

PR #22, opened 2021-07-15
translate

jongbonga:


web-client/src/app/views/landing/landing.page.html
```diff
@@ -15,18 +15,19 @@ <h1 class="ion-text-center font-black">NAUTILUS</h1>
...
           fill="outline"
           routerLink="/wallet"
-          >OPEN WALLET
-        </ion-button>
+          transloco="LANDING.open"
+        ></ion-button>
       </ion-col>
       <ion-col size="12">
-        <p class="ion-text-center">Terms & conditions apply</p>
+        <p class="ion-text-center">Terms & Conditions apply</p>
```
PiDelport, line 30: Should we get the rest of this text imported to Transloco too?

web-client/src/app/transloco/transloco-root.module.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+@NgModule({
+  exports: [TranslocoModule],
+  providers: [
+    {
+      provide: TRANSLOCO_CONFIG,
+      useValue: translocoConfig({
+        availableLangs: ['en', 'fr'],
+        defaultLang: 'en',
```
PiDelport, line 29: Should we default to the browser language here?

```suggestion
        defaultLang: getBrowserLang(),
        fallbackLang: 'en',
```

web-client/src/app/transloco/transloco-root.module.ts
```diff
@@ -0,0 +1,38 @@
+import { HttpClient } from '@angular/common/http';
+import { Injectable, NgModule } from '@angular/core';
+import {
```
PiDelport, line 3: (Import for below)

```suggestion
import {
  getBrowserLang,
```

PiDelport (approved):

web-client/src/app/views/landing/landing.page.html
jongbonga: Yes, I'm planning on doing it when this branch is merged. The goal for the PR is mostly to set the plugin.

web-client/src/app/transloco/transloco-root.module.ts
jongbonga: We still going to need translation files for each language. setting this up right now will just bring unnecessary problems

PiDelport: With `fallbackLang` set, it should still be fine, right?

jongbonga: Yes, but there is no need to do it if we know for a fact that it will fall back anyway!
We will implement when we have at least 1 additional language fully translated.

(I will do fr after the UI is solid)

PiDelport: I think it's worth doing now: it makes it easy to spot-test the localisation by changing the browser's locale. (That's how I tested this PR.)

We'll want to be able to do this kind of quick interactive switching while we develop the rest of the app, to make sure fields are correctly tagged, and to test partial translations; this should be the most low-cost and low-risk way of doing that. (The alternative would be having to change the default in the source code and recompiling every time we want to switch it, but that's a lot worse, I think.)

Since this is something we'll need anyway, and is already useful for development and testing right now, I don't think there's a reason to delay it, is there?

jongbonga: >I don't think there's a reason to delay it, is there?

You are totally right, but there is no reason to buy paints if there are no walls yet. 
When we start translating (which will happen @ some point), we will implement it :).






> The alternative would be having to change the default in the source code and recompiling every time we want to switch it

There is no recompiling with this plugin, that's the #1 reason I went with it over the built-in Angular i18n. 
The translation is done JIT

PiDelport: No, I mean having to recompile to switch the actual UI for previewing, since we don't have other UI for switching the app's current language dynamically.

Using the browser preference just means we don't have to implement an in-app UI (or change the source default and recompile) to be able to do that.

PiDelport:
Note: The 357afc9 merge actually rolled back the required `package-lock.json` changes.

I'm not sure why `npm ci` didn't catch that: it's supposed to.

PiDelport:
It looks like that's actually a `npm ci` bug / regression:

- https://github.com/npm/cli/issues/2701

web-client/src/app/transloco/transloco-root.module.ts
PiDelport: (Resolved in #32.)

web-client/src/app/transloco/transloco-root.module.ts
PiDelport: (Resolved in #32.)

jongbonga: > * Resolved in: [translate 2 #32](https://github.com/registreerocks/nautilus-wallet/pull/32)

You just had to do it :) ...

PR #23, opened 2021-07-18
perf(web-client): set global max-width & bg-color

jongbonga:


PiDelport (changes requested):
We should apply these styles to Storybook too.

Something like this should do it?

```diff
--- a/web-client/.storybook/preview.js
+++ b/web-client/.storybook/preview.js
@@ -18,6 +18,10 @@ export const parameters = {
   docs: { inlineStories: true },
 };
 
+// Apply AppComponent's style to all stories.
+// Docs: https://storybook.js.org/docs/angular/configure/styling-and-css#importing-css-files
+import("!style-loader!css-loader!../src/app/app.component.scss");
+
 // Provide context for Ionic components
 export const decorators = [
   // See AppModule:
@@ -28,7 +32,7 @@ export const decorators = [
   componentWrapperDecorator(
     (story) => `
       <ion-app>
-        <ion-router-outlet>${story}</ion-router-outlet>
+        <ion-router-outlet class="max-w-screen-sm m-auto">${story}</ion-router-outlet>
       </ion-app>
     `
   ),
```

(We could avoid having to modify the template too by moving those class's styles into `app.component.scss` too?)

PiDelport (changes requested):
We should still add the style import in `.storybook/preview.js`, so the page stories render correctly?

longtomjr (approved):
Happy to have this merged once all the feedback have been addressed

jongbonga:
> We should still add the style import in `.storybook/preview.js`, so the page stories render correctly?

Why? 
It's rendering correctly on my side.

PiDelport:
> It's rendering correctly on my side.

The styles are not being applied to the Storybook stories:

- https://60ddb6c36599fd003b211fd8-bocbssrdjt.chromatic.com/?path=/story/views-landingpage--default
- https://www.chromatic.com/test?appId=60ddb6c36599fd003b211fd8&id=60f8a04277fd2700390aa054

To fix that, we need to import the style sheet in Storybook:

```diff
   docs: { inlineStories: true },
 };
 
+// Apply AppComponent's style to all stories.
+// Docs: https://storybook.js.org/docs/angular/configure/styling-and-css#importing-css-files
+import("!style-loader!css-loader!../src/app/app.component.scss");
+
 // Provide context for Ionic components
 export const decorators = [
   // See AppModule:
```

Nghondzweni (approved):

web-client/src/global.scss
```diff
@@ -24,3 +24,10 @@
 @import "~@ionic/angular/css/text-alignment.css";
 @import "~@ionic/angular/css/text-transformation.css";
 @import "~@ionic/angular/css/flex-utils.css";
+
+ion-toolbar,
+ion-content,
+ion-footer {
+  --background: var(--ion-color-prussian-blue);
+  --color: var(--ion-color-white);
+}
```
PiDelport, line 33: This creates white margins on either side of the app: [see UI change](https://www.chromatic.com/pullrequest?appId=60ddb6c36599fd003b211fd8&number=23&view=changes)

Rather than targeting these individual Ionic components, we should probably just set Ionic's global style variables, which should take effect across all components, like this?

```suggestion
:root {
  // Docs: https://ionicframework.com/docs/theming/themes#application-colors
  --ion-background-color: var(--ion-color-prussian-blue);
  --ion-background-color-rgb: var(--ion-color-prussian-blue-rgb);
  --ion-text-color: var(--ion-color-white);
  --ion-text-color-rgb: var(--ion-color-white-rgb);
}
```

PiDelport (changes requested):

PiDelport:
Cool, the [UI change](https://www.chromatic.com/pullrequest?appId=60ddb6c36599fd003b211fd8&number=23&view=changes) looks good now!

PiDelport (approved):
Cool, the [UI changes](https://www.chromatic.com/pullrequest?appId=60ddb6c36599fd003b211fd8&number=23&view=changes) look good now!

PR #24, opened 2021-07-18
header-component

jongbonga:


web-client/src/app/components/header/header.component.html
```diff
@@ -0,0 +1,23 @@
+<ion-toolbar>
+  <ion-buttons slot="start">
+    <ion-button>
+      <ion-icon name="menu" src="icon-only"></ion-icon>
+    </ion-button>
+  </ion-buttons>
+  <ion-grid>
+    <ion-row class="ion-align-items-center">
+      <ion-col size="auto">
+        <img src="assets/img/Nautilus.svg" />
```
PiDelport, line 10: It's probably better to import and use the Nasalisation font for this, rather than SVG?

web-client/src/app/components/header/header.component.html
```diff
@@ -0,0 +1,23 @@
...
+      </ion-col>
+      <ion-col>
+        <ion-button
+          size="small"
+          expand="block"
+          shape="round"
+          color="portland-orange"
+          >DEPOSIT
+        </ion-button>
```
PiDelport, line 19: I don't think we actually want a button here. The initial mock-ups have one, but see [this Figma thread](https://www.figma.com/file/zYVHu1EwCIDCbQp8rFscyA?node-id=39:322#94660054): let's talk about that before committing here?

PiDelport (changes requested):
We should probably also wait for the Storybook style changes from:

- #23

web-client/src/app/components/header/header.component.html
jongbonga: The title isn't using Nasalisation. it's a custom vector

longtomjr (approved):
Happy to get this merged once all feedback have been addressed.We don't need to worry too much about which components goes where at this point, since the designs are going to keep evolving.

web-client/src/app/components/header/header.component.html
PiDelport: Ah, did Abby draw it from scratch, or is it a vector export of some other source?

jongbonga: I think she did, I just exported from the design.

PiDelport: Okay, I checked with Abby: she says it's actually this font:

- https://fonts.google.com/specimen/Audiowide

Can we import and use that?

web-client/src/app/components/header/header.component.html
PiDelport: (I'm in favour of just leaving this out for now, rather than adding and having to remove it again later.)

web-client/src/app/components/header/header.component.html
jongbonga: I imported it in #31 and replaced the image with text.

PiDelport:
- Depends on #31

(Marking this as draft until that's in.)

Nghondzweni (approved):

jongbonga:
@PiDelport can we look into this too (after #23) since its blocker #31 is in?
#30 depends on it

PiDelport (approved):
Awesome, looks good!

PR #28, opened 2021-07-21
pref(web-client): update theme colors

jongbonga:
- Follows #23


longtomjr (approved):
Looks good. I assume the colors align with the designs.

jongbonga:
> Looks good. I assume the colors align with the designs.

Yes

web-client/src/theme/variables.scss
```diff
@@ -8,69 +8,68 @@
 
 /** Ionic CSS Variables **/
 :root {
```
PiDelport, line 10: We should probably link the UI Kit here, for easy reference while maintaining:

```suggestion
:root {
  // See Nautilus UI Kit:
  // https://www.figma.com/file/zYVHu1EwCIDCbQp8rFscyA/Nautilus-Branding?node-id=23%3A128

```

web-client/src/theme/variables.scss
```diff
@@ -8,69 +8,68 @@
...
-  /** primary **/
-  --ion-color-primary: #3880ff;
-  --ion-color-primary-rgb: 56, 128, 255;
+  --ion-color-primary: #003450;
```
PiDelport, line 11:
```suggestion
  // UI Kit: Primary 1 - Prussian Blue
  --ion-color-primary: #003450;
```

PiDelport, line 11: I think it would be good to comment these colours with which UI Kit role / colour they correspond to, for easier maintenance.

I made a changeset for that while checking the colours: @jongbonga, can I add that for review?

PiDelport (changes requested):
Cool, just some minor feedback.

- Note that we'll have to resolve this with #23, depending on which gets merged first.

web-client/src/theme/variables.scss
jongbonga: Yes

PiDelport:
- #32 should fix the UI test regression (marking this as draft until that's in)

jongbonga:
> * [translate 2 #32](https://github.com/registreerocks/nautilus-wallet/pull/32) should fix the UI test regression (marking this as draft until that's in)

what do theme colors have to do with translation?

PiDelport:
> what does theme colors have to do with translation?

Well, we merged #22 a bit eagerly (my fault 😅), which broke the baseline for the landing page UI test, causing [this regression (Build 62)](https://www.chromatic.com/test?appId=60ddb6c36599fd003b211fd8&id=60f94180105290003b26ca4d) in this PR.

PR #32 fixed the baseline, so now we can see the effect of this PR without the regression above: [Build 65](https://www.chromatic.com/test?appId=60ddb6c36599fd003b211fd8&id=60f95553629f7f00394dc344)

web-client/tailwind.config.js
```diff
@@ -5,17 +5,17 @@ module.exports = {
...
-        "prussian-blue": "var(--ion-color-prussian-blue)",
-        "usafa-blue": "var(--ion-color-usafa-blue)",
-        "viridian-green": "var(--ion-color-viridian-green)",
-        "portland-orange": "var(--ion-color-portland-orange)",
-        "blue-jeans": "var(--ion-color-blue-jeans)",
-        gainsboro: "var(--ion-color-gainsboro)",
-        "caribbean-green": "var(--ion-color-caribbean-green)",
-        mustard: "var(--ion-color-mustard)",
+        primary: "var(--ion-color-primary)",
+        seconday: "var(--ion-color-seconday)",
```
PiDelport, line 9:
```suggestion
        secondary: "var(--ion-color-secondary)",
```

PiDelport (changes requested):
Okay, we should just update the removed color reference in `landing.page.html` too:

```diff
--- a/web-client/src/app/views/landing/landing.page.html
+++ b/web-client/src/app/views/landing/landing.page.html
@@ -1,4 +1,4 @@
-<ion-content color="prussian-blue">
+<ion-content color="primary">
   <ion-grid class="h-full sm:max-w-max">
     <ion-row class="h-full content-around px-5">
       <ion-col size="12">
```

This should fix the remaining UI test too.

- Alternatively, we can merge #23 first, and update this?

jongbonga:
> we should just update the removed color reference in `landing.page.html` too

This change can be done on a different branch.
Let's not involve pages in this PR

PiDelport:
> This change can be done on a different branch.
> Let's not involve pages in this PR

Well, the color name is what this branch is about: if we remove the `prussian-blue` color name in this PR, we should also update the references to it in this PR, to keep `main` in a working state. Otherwise, the UI tests for all the open PRs will be in a bad / failing state.

jongbonga:
> > This change can be done on a different branch.
> > Let's not involve pages in this PR
> 
> Well, the color name is what this branch is about: if we remove the `prussian-blue` color name in this PR, we should also update the references to it in this PR, to keep `main` in a working state. Otherwise, the UI tests for all the open PRs will be in a bad / failing state.

Can I be trusted? if yes, please let me deal with it.
If no, then do what you think is best.

jongbonga:
@PiDelport removed the color from the content since it's being applying globally on #23

Nghondzweni (approved):

PiDelport (changes requested):

web-client/src/app/views/landing/landing.page.html
```diff
@@ -1,5 +1,5 @@
-<ion-content color="prussian-blue">
-  <ion-grid class="h-full sm:max-w-max">
+<ion-content>
+  <ion-grid class="h-full">
```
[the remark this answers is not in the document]
jongbonga: This won't be necessary and will probably need to be removed when #23 get merged because the content color is set globally & the `max-width` is also being applied on `router-outlet` level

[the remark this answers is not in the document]
PiDelport: Right, #23 will change this, but we should still merge a version with the right color in this PR, so that the baseline snapshots in Chromatic looks correct again. (The baseline snapshot is what all the following PRs will be compared with, so it's worth getting right, so that the UI tests can catch unintentional regressions.)

Does that sound good?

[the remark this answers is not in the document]
jongbonga: Then let's merge #23 first so we have to change this one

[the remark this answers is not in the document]
PiDelport: Cool, sounds good.

PiDelport (changes requested):
Cool, this just needs updating in `global.scss` now:

```diff

 :root {
   // Docs: https://ionicframework.com/docs/theming/themes#application-colors
-  --ion-background-color: var(--ion-color-prussian-blue);
-  --ion-background-color-rgb: var(--ion-color-prussian-blue-rgb);
+  --ion-background-color: var(--ion-color-primary);
+  --ion-background-color-rgb: var(--ion-color-primary-rgb);
   --ion-text-color: var(--ion-color-white);
   --ion-text-color-rgb: var(--ion-color-white-rgb);
 }
```

web-client/src/global.scss
```diff
@@ -27,8 +27,8 @@
 
 :root {
   // Docs: https://ionicframework.com/docs/theming/themes#application-colors
-  --ion-background-color: var(--ion-color-prussian-blue);
-  --ion-background-color-rgb: var(--ion-color-prussian-blue-rgb);
+  --ion-background-color: var(--ion-color-primary);
+  --ion-background-color-rgb: var(--ion-color-primary-rgb);
   --ion-text-color: var(--ion-color-white);
   --ion-text-color-rgb: var(--ion-color-white-rgb);
 }
```
PiDelport, line 34: We should update Ionic's stepped colours here too:

(Ionic relies on these being sync with the background and text colours: see the [Stepped Colors docs](https://ionicframework.com/docs/theming/themes#stepped-colors))

```suggestion

  // Generated with: https://ionicframework.com/docs/theming/themes#stepped-color-generator
  // NOTE: Re-generate these whenever --ion-background-color / --ion-text-color change
  --ion-color-step-50: #0d3e59;
  --ion-color-step-100: #1a4862;
  --ion-color-step-150: #26526a;
  --ion-color-step-200: #335d73;
  --ion-color-step-250: #40677c;
  --ion-color-step-300: #4d7185;
  --ion-color-step-350: #597b8d;
  --ion-color-step-400: #668596;
  --ion-color-step-450: #738f9f;
  --ion-color-step-500: #809aa8;
  --ion-color-step-550: #8ca4b0;
  --ion-color-step-600: #99aeb9;
  --ion-color-step-650: #a6b8c2;
  --ion-color-step-700: #b3c2cb;
  --ion-color-step-750: #bfccd3;
  --ion-color-step-800: #ccd6dc;
  --ion-color-step-850: #d9e1e5;
  --ion-color-step-900: #e6ebee;
  --ion-color-step-950: #f2f5f6;
}
```

web-client/src/global.scss
```diff
@@ -27,8 +27,8 @@
 
 :root {
   // Docs: https://ionicframework.com/docs/theming/themes#application-colors
-  --ion-background-color: var(--ion-color-prussian-blue);
-  --ion-background-color-rgb: var(--ion-color-prussian-blue-rgb);
+  --ion-background-color: var(--ion-color-primary);
+  --ion-background-color-rgb: var(--ion-color-primary-rgb);
```
PiDelport, line 31: We should rename this to avoid a conflict between the background and primary colors in Ionic's color system: Ionic expects `primary` to be usable on top of the background color. (If they're the same, a lot of Ionic's assumptions break down, and will render wrong.)

For the Nautilus theme, the Ionic `primary` should probably be defined to be Portland Orange, and the background (Prussian Blue) should be named something other than `primary`, such as `primary-background`.

This diff implements that suggestions: cf254e2 – how does that look?

PiDelport (changes requested):
Cool, with these two things addressed, it should be good to go!

web-client/src/global.scss
jongbonga: I'm more comfortable with implementing this on a different PR.

longtomjr: #33

PiDelport: I think it makes more sense to fix this in this PR?

I added a story and pushed some temporary branches to demonstrate the issue more clearly:

1. [Baseline](https://www.chromatic.com/test?appId=60ddb6c36599fd003b211fd8&id=60fe7994ca4d240039ee58de) - How the theme looks like on main (with the stepped color fix from this branch).
2. [Without fix](https://www.chromatic.com/test?appId=60ddb6c36599fd003b211fd8&id=60fe7afcca4d240039ee94bd) - How the branch currently looks: in particular, see how the primary / background color conflict doesn't just affect the default button styling, but makes outline buttons and links completely invisible. (It probably breaks more than this: every Ionic component that references the primary color will render unexpectedly.)
3. [With fix](https://www.chromatic.com/test?appId=60ddb6c36599fd003b211fd8&id=60fe7b00db358200391164c9) - How this branch looks with the suggested fix applied.

We don't necessarily have to use that particular fix, as long as we do something that doesn't break the Ionic color system's assumptions?

PiDelport (commented):
Looks good! 🚀

PR #34, opened 2021-07-26
theme-color-quick-fix

jongbonga:
This fixes the primary contrast's color which was previously defaulted to #000000, it also updates tailwind colors link


web-client/src/theme/variables.scss
```diff
@@ -32,7 +32,7 @@
   // UI Kit: Primary 2 - Portland Orange
   --ion-color-primary: #ff642e;
   --ion-color-primary-rgb: 255, 100, 46;
-  --ion-color-primary-contrast: #000000;
+  --ion-color-primary-contrast: #ffffff;
```
PiDelport, line 35: Is there a reason for deviating from the Ionic color generator's contrast color here?

If so, we should document it?

web-client/src/theme/variables.scss
```diff
@@ -32,7 +32,7 @@
...
   --ion-color-primary-contrast-rgb: 0, 0, 0;
```
PiDelport, line 36: (The `contrast` and `contrast_rgb` variations should match.)

PiDelport (changes requested):

web-client/src/theme/variables.scss
longtomjr: This contrast match the designs, so yep, we should document it for now. Also not sure what the contrast ratio is here, if it is not great we should raise this with Abby

web-client/src/theme/variables.scss
jongbonga: Done! Thanks

PiDelport (approved):
Looks good! 🚀

PR #35, opened 2021-07-26
test(web-client): add story for the Ionic theme

PiDelport:
- From #28

This story gives us a quick view of how the Nautilus brand palette maps to the Ionic theme.

(This could also be a starting point for a [design system](https://storybook.js.org/tutorials/design-systems-for-developers/) for the Nautilus brand.)

### DIFF
--- a/web-client/src/stories/theme.stories.ts
+++ b/web-client/src/stories/theme.stories.ts
@@ -1,0 +1,12 @@
+import { Component } from '@angular/core';
+import { Meta, Story } from '@storybook/angular';
+
+export default {
+  title: 'Theme',
+} as Meta;
+
+export const Theme: Story<void> = () => ThemeComponent;
+
+const ThemeComponent: Component = {
+  templateUrl: './theme.html',
+};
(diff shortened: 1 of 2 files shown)


web-client/src/stories/theme.stories.ts
```diff
@@ -0,0 +1,9 @@
+import { Meta } from '@storybook/angular';
+
+export default {
+  title: 'Theme',
+} as Meta;
+
+export const Theme = () => ({
```
jongbonga, line 7: GitHub Actions/test is complaining about a missing return type on function

jongbonga (changes requested):

web-client/src/stories/theme.stories.ts
PiDelport: Good catch!

[c6c26b9..6c48d84](https://github.com/registreerocks/nautilus-wallet/compare/c6c26b9d8e91a25b99c46d8287b6ab4afbcec82b..6c48d8466958e5ff17a4264d77be3d28b5615520) should resolve it: how does that look?

jongbonga (approved):
Looks good

PR #36, opened 2021-07-27
test(web-client): add basic coverage for routes / navigation

PiDelport:
- From #30

### DIFF
--- a/web-client/.storybook/tsconfig.json
+++ b/web-client/.storybook/tsconfig.json
@@ -4,6 +4,7 @@
     "types": ["node"]
   },
   "exclude": [
+    "../src/tests",
     "../src/test.ts",
     "../src/**/*.spec.ts",
     "../projects/**/*.spec.ts"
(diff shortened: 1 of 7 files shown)


PiDelport:
Note: The coverage drop here is actually due to https://github.com/angular/angular-cli/issues/1735.

- #37 should work around it.

jongbonga (approved):

PiDelport:
Alright, as far as I can tell, the intermittent test failure was actually a secondary error due to an Angular module consistency error, due to the dynamic test module's component declaration conflicting with the code's actual module's declaration. The message from this error doesn't get reported reliably, because it happens as part of an uncaught promise failure, due to the default Angular test code still using `waitForAsync` instead of more reliable `async` / `await` patterns, apparently.

Apparently Angular 12 is trying to be better about this: 

* https://stackoverflow.com/questions/64895893/why-testbed-beforeeach-uses-waitforasync-instead-of-just-async-await-compilecomp#comment119275081_64897437
* Angular docs issue: https://github.com/angular/angular/issues/39740
  * Recently fixed in: https://github.com/angular/angular/commit/2b939767fe3a43eb6b3eb220a5b8670ff66bbaec

I didn't commit the `waitForAsync` → `async`/`await` changes I temporarily made to get the error message more reliably—that's probably something we should do more holistically for all the tests, at some point—but 22a3503 should fix the Angular module consistency issue and make this branch pass, at least.

PR #37, opened 2021-07-27
test(web-client): fix coverage reporting

PiDelport:
This works around upstream Angular issue:

https://github.com/angular/angular-cli/issues/1735

This will decrease the reported coverage percentage, but the adjusted number should actually be correct now.

### DIFF
--- a/web-client/src/test.ts
+++ b/web-client/src/test.ts
@@ -19,13 +19,26 @@ declare const require: {
     <T>(id: string): T;
   };
 };
-
 // First, initialize the Angular testing environment.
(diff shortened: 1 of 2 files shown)


jongbonga (approved):

PiDelport:
Hmm, I'm not sure why the UI Tests status check are still showing as pending here on GitHub: they're passed on Chromatic.

May be a GitHub integration niggle: I'll trigger a rebuild to see if that helps?

PiDelport:
- #40 should fix the E2E tests.

PR #39, opened 2021-07-28
Wallet

jongbonga:


longtomjr (approved):
Looks good.

web-client/src/app/views/wallet/wallet.page.spec.ts
```diff
@@ -15,12 +15,12 @@ describe('WalletPage', () => {
   beforeEach(
     waitForAsync(() => {
       TestBed.configureTestingModule({
-        declarations: [],
+        declarations: [WalletPage],
```
PiDelport, line 18: Hmm, this undoes the fix in #36, and creates a conflicting declaration again.

It looks like the tests still didn't catch this reliably: when running `npm test` in watch mode locally, it seems to fail more reliably after the first run, but the initial runs still pass some of the time depending on test order or other timing, it looks like.

I took a stab at making this more reliable in:

- #44

web-client/src/app/views/wallet/wallet-routing.module.ts
```diff
@@ -2,7 +2,7 @@ import { NgModule } from '@angular/core';
 import { RouterModule, Routes } from '@angular/router';
 import { WalletPage } from './wallet.page';
 
-const routes: Routes = [
+export const routes: Routes = [
```
PiDelport, line 5: (This should probably stay internal?)

PR #40, opened 2021-07-28
ci(web-client): disable ng e2e's webdriverUpdate behaviour for GitHub Actions

PiDelport:
This should avoid the e2e tests breaking whenever the GitHub Actions
virtual environment lags behind the latest Chrome major version release.

### DIFF
--- a/.github/workflows/web-client.yaml
+++ b/.github/workflows/web-client.yaml
@@ -1,6 +1,7 @@
 on:
   push:
     paths:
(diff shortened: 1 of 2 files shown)


longtomjr (approved):
Lets get this in

PiDelport:
Alright, 691d7e5 should the last niggle, and this should actually be reliable now! (Or at least more reliable than before!) 🎉

PR #41, opened 2021-07-28
feat(web-client): add fontawesome

jongbonga:


longtomjr (approved):
Looks good.

longtomjr:
@jongbonga Prettier check is failing, just check that on your end please

jongbonga:
> @jongbonga Prettier check is failing, just check that on your end please

Yes, just pushed an update for that.

longtomjr (changes requested):
For some reason there is a UI change in chromatic that is broken. Please have a look at that @jongbonga

longtomjr (approved):
Nevermind, the issue is the wrong baseline that were chosen by chromatic.  Will update the baseline there and then we can get this merged.

PR #45, opened 2021-08-04
receive

jongbonga:


web-client/src/app/views/receive/receive.page.html
```diff
@@ -0,0 +1,33 @@
...
+        elementType="img"
+        [width]="1000"
+        [margin]="2"
+      ></qrcode>
+    </ion-card-content>
+  </ion-card>
+  <ion-list>
+    <app-action-item
+      *ngFor="let item of actionItems"
+      [preTitle]="item.preTitle"
```
longtomjr, line 28: We just need to check the styling for the preTitle here. The contrast seems a bit low, not sure if that is how it is on the designs. See the UI Review in chromatic.

longtomjr (commented):
Just a quick comment on the preTitle styling.

longtomjr (approved):
Look good thanks!

PR #46, opened 2021-08-12
build(web-client): install & setup akita

jongbonga:


web-client/package.json
```diff
@@ -31,6 +36,10 @@
     "@capacitor/haptics": "1.0.2",
     "@capacitor/keyboard": "1.0.2",
     "@capacitor/status-bar": "1.0.2",
+    "@datorama/akita": "^6.2.3",
+    "@datorama/akita-cli": "^6.1.1",
```
longtomjr, line 40: Can the CLI (and the devtools below) be moved to dev-dependencies?

jongbonga: Done

PR #48, opened 2021-08-12
feat(web-server): port crypto and attestation code

PiDelport:
- Issue: #1

PRs:

- Follows #47
- Precedes #49

### DIFF
--- a/web-server/sgx-wallet-test/enclave/src/ported/proptest_crypto.rs
+++ b/web-server/sgx-wallet-test/enclave/src/ported/proptest_crypto.rs
@@ -23,7 +23,7 @@ fn prop_soda_box_roundtrips_impl(
     assert_ne!(sender.get_pubkey(), receiver.get_pubkey());
 
     let encrypted = sender
-        .encrypt_message(Secret::new(plaintext.clone()), &receiver.get_pubkey())
+        .encrypt_message(&Secret::new(plaintext.clone()), &receiver.get_pubkey())
         .unwrap();
     let decrypted = receiver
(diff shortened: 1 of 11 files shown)


web-server/sgx-wallet-test/enclave/src/ported/proptest_crypto.rs
```diff
@@ -0,0 +1,37 @@
...
+        .unwrap();
+    let decrypted = receiver
+        .decrypt_message(
+            &encrypted.ciphertext,
+            &sender.get_pubkey(),
+            &encrypted.nonce,
+        )
+        .unwrap();
+
+    assert_eq!(decrypted.expose_secret(), &plaintext);
```
longtomjr, line 36: For assertions related to the test we might want to use the proptest assert macros:
https://altsysrq.github.io/rustdoc/proptest/latest/proptest/macro.prop_assert.html
https://altsysrq.github.io/rustdoc/proptest/latest/proptest/macro.prop_assert_eq.html

It does require the functions to return a `Result<_, TestCaseError>` instead since it does not panic, but I think that should be fine.

longtomjr (approved):
Looks good. Just the one maybe change with the proptest asserts

web-server/sgx-wallet-test/enclave/src/ported/proptest_crypto.rs
PiDelport: Thanks for spotting! We can't use the closure style directly, but we can use the wrapper: fc798e6

PR #49, opened 2021-08-12
feat(web-server): port SGX KvStore code

### LINKED ISSUE #4: Pull out KvStore functionality from trusted compute (separate crate)
PRs:

- [ ] #49

PiDelport:
- Closes #4

PRs:

- Follows #48
- Precedes #67

### DIFF
--- a/web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
+++ b/web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
@@ -48,7 +48,7 @@ pub trait KvStore<V> {
 
     /// Mutate the value of `key`.
     ///
-    /// This is like [`Self::mutate`], but only operates on existing values.
+    /// This is like [`Self::alter`], but only operates on existing values.
     fn mutate<F>(&mut self, key: &Key, mutate_fn: F) -> Result<Option<V>, Self::Error>
     where
         F: FnOnce(V) -> V,
--- a/web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
+++ b/web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
@@ -1,0 +1,69 @@
+//! Simple key-value store abstraction
+//!
+//! XXX port note: Key type changed from str to `[u8]`.
+
+pub mod fs;
+
+pub type Key = [u8];
+
+/// A key-value store.
+///
+/// These methods borrow key and value references,
+/// to suit cloning / serialising implementations.
+///
+pub trait KvStore<V> {
+    type Error;
+
+    /// Load the saved value for `key`, if any.
+    ///
+    /// Return [`None`] if `key` has no previous value.
+    ///
+    fn load(&self, key: &Key) -> Result<Option<V>, Self::Error>;
+
+    /// Save a new value for `key`.
+    ///
+    /// This will replace any existing value.
+    ///
+    fn save(&mut self, key: &Key, value: &V) -> Result<(), Self::Error>;
+
+    /// Delete the saved value for `key`.
+    fn delete(&mut self, key: &Key) -> Result<(), Self::Error>;
+
+    /// Alter the value of `key`.
+    ///
+    /// This operation is a generalisation of [`Self::load`], [`Self::save`], and [`Self::delete`].
+    ///
+    fn alter<F>(&mut self, key: &Key, alter_fn: F) -> Result<Option<V>, Self::Error>
+    where
+        F: FnOnce(Option<V>) -> Option<V>,
+    {
+        let loaded: Option<V> = self.load(key)?;
+        let altered: Option<V> = alter_fn(loaded);
+        match &altered {
+            None => self.delete(key)?,
+            Some(value) => self.save(key, value)?,
+        };
+        Ok(altered)
+    }
+
+    /// Mutate the value of `key`.
+    ///
+    /// This is like [`Self::mutate`], but only operates on existing values.
+    fn mutate<F>(&mut self, key: &Key, mutate_fn: F) -> Result<Option<V>, Self::Error>
+    where
+        F: FnOnce(V) -> V,
+    {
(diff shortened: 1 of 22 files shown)


web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
```diff
@@ -0,0 +1,69 @@
+//! Simple key-value store abstraction
+//!
+//! XXX port note: Key type changed from str to `[u8]`.
```
longtomjr, line 3:
```suggestion
//! XXX port notes: 
//!   - Key type changed from str to `[u8]`.
//!   - Added `KvStore::mutate`
//!   - Added `KvStore::try_insert`
```

web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+        match &altered {
+            None => self.delete(key)?,
+            Some(value) => self.save(key, value)?,
+        };
+        Ok(altered)
+    }
+
+    /// Mutate the value of `key`.
+    ///
+    /// This is like [`Self::mutate`], but only operates on existing values.
```
longtomjr, line 51:
```suggestion
    /// This is like [`Self::alter`], but only operates on existing values.
```

web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+    fn mutate<F>(&mut self, key: &Key, mutate_fn: F) -> Result<Option<V>, Self::Error>
+    where
+        F: FnOnce(V) -> V,
```
longtomjr, line 54: Maybe `mutate_fn` should implement `FnOnce(V) -> Option<V>` to cater for deletion. The other option is to just not allow deletion, document it, and enforce it with the current signature.

web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+            None => self.delete(key)?,
+            Some(value) => self.save(key, value)?,
+        };
+        Ok(altered)
+    }
+
+    /// Mutate the value of `key`.
+    ///
+    /// This is like [`Self::mutate`], but only operates on existing values.
+    fn mutate<F>(&mut self, key: &Key, mutate_fn: F) -> Result<Option<V>, Self::Error>
```
longtomjr, line 52: Signature wise, it might make sense for this to return `Result<V, Self::Error>` instead, depending on if we cater for deletion or not. If I am reading this correctly, this would return None in when trying to mutate a non existing keypair. I think in that case an Error might be more predictable.

I think it helps to be clear about mutations failing in this context.

web-server/sgx-wallet-test/enclave/src/ported/test_kv_store_fs.rs
```diff
@@ -0,0 +1,33 @@
...
+use sgx_wallet_impl::ported::kv_store::fs::{decode_from_fs_safe, encode_to_fs_safe};
+use sgx_wallet_impl::ported::kv_store::Key;
+
+/// [`encode_to_fs_safe`] encodes to filesystem-safe, and [`decode_from_fs_safe`] round-trips.
+pub(crate) fn prop_fs_safe_roundtrip() {
+    prop_fs_safe_roundtrip_wrapper()
+}
+
+proptest! {
+    fn prop_fs_safe_roundtrip_wrapper(key: Box<Key>) {
```
longtomjr, line 13: Can we use the closure style invocation here as well to stay consistent inside of the project or will it not work for this case?

PiDelport: It won't work for this case currently: the closure-style invocation doesn't support specifying the return type, which is necessary for using `prop_assert!` (which is a hard requirement for the enclave-based tests, because `assert!` panics are actually fatal / not caught by proptest).

(I have some in-progress project documentation for this, which I'll follow-up with at some point.)

web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
PiDelport: The way the `Self::Error` type is parameterised is intended more for exceptional failure conditions (particularly IO), so we can't really use it here.

We could introduce a separate, but I don't think that gives us much semantic advantage over returning the result of the operation directly as an `Option`: this is more consistent with how the rest of the API operates.

(We could probably document it better later, but the intent of this method is not "mutate and assert that the key was present", but rather "mutate if it was present, and return whether it was mutated (along with the result, if it was)".)

PiDelport: (Resolving this for now: we should probably defer bigger changes to this ported code until we've split it out to a shared library.)

web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
PiDelport: We can probably just use `alter`, for that case?

PiDelport: (Resolving this for now: we should probably defer bigger changes to this ported code until we've split it out to a shared library.)

web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
PiDelport: `mutate` and `try_insert` are part of the ported code, not port additions, so we probably don't need this?

Nghondzweni (approved):

web-server/sgx-wallet-impl/src/ported/kv_store/mod.rs
longtomjr: Ah did not realize you were porting from a different branch than main.

PR #50, opened 2021-08-16
Scan qrcode

jongbonga:


web-client/src/app/services/scanner.service.ts
```diff
@@ -0,0 +1,37 @@
...
+  async requestPermissions() {
+    //NOTE: https://github.com/ionic-team/capacitor/discussions/4944#discussioncomment-1205023
+    try {
+      //NOTE: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia
+      const { active } = await navigator.mediaDevices.getUserMedia({
+        video: true,
+      });
+      return active;
+    } catch (e) {
+      console.log(e);
```
longtomjr, line 33: Extra console log.

web-client/src/app/views/scanner/scanner.page.ts
```diff
@@ -0,0 +1,25 @@
...
+  templateUrl: './scanner.page.html',
+  styleUrls: ['./scanner.page.scss'],
+})
+export class ScannerPage implements OnInit {
+  constructor(private modalCtrl: ModalController) {}
+
+  ngOnInit() {}
+
+  scanSuccessHandler(event: any) {
+    console.log(event);
```
longtomjr, line 15: I assume this is still to be implemented in a follow up PR. Is that correct?

web-client/src/app/views/send-funds/send-funds.page.ts
```diff
@@ -12,15 +13,32 @@ export class SendFundsPage implements OnInit {
...
-  constructor() {}
+  constructor(private scannerService: ScannerService) {}
 
   ngOnInit() {}
+
+  async presentScanner() {
+    await this.scannerService.presentScanner();
+  }
+
+  perform(action: string | undefined) {
```
longtomjr, line 34: Maybe rename this function to "perform_action" or something that describes what the function does in a clearer way.

longtomjr (changes requested):
A couple things. We also need to get the test coverage up before merging. Not sure if it makes sense to cover all the functions at this point but we can chat about that tomorrow in our call.

web-client/src/app/views/scanner/scanner.page.ts
jongbonga: Yes

longtomjr: I think adding a `// TODO` comment so it is clear that this is not implemented yet when reviewing and working with the code in the future will help for cases like this.

longtomjr (approved):
Happy for this to go in once we have the test coverage sorted. Will create and pull changes into a separate branch from this one for the demo for now.

web-client/src/app/services/scanner.service.ts
```diff
@@ -7,6 +7,7 @@ import { Camera, CameraPermissionState } from '@capacitor/camera';
 export class ScannerService {
   constructor() {}
 
+  /* istanbul ignore next not sure if there is a way to easily spy on Camera.checkPermissions() */
```
jongbonga, line 10: I saw [this](https://capacitorjs.com/docs/guides/mocking-plugins) yesterday, not sure if it's relevant

longtomjr: Aaah, that seems to be the way to do it. I think it make sense to have this go in separately for now. Created an issue #62

PR #52, opened 2021-08-19
refactor(web-client): replace text with logo img

jongbonga:


longtomjr:
As a note. We had a call about this, and this seems to be the best way to have the text not overflow on screens without using a plugin. I think that we want the logo + text as an SVG in anyways to preserve the spacing and future elements that might deviate from the font.

longtomjr (approved):

PR #54, opened 2021-08-19
Add registration page

Nghondzweni:
I set the pull request to draft, there are still a bit of changes that I want to make. But overall the page is relatively complete.


longtomjr (changes requested):
There seems to be a massive package-lock change. It might be a npm version issue on your end. We will just need to resolve that before merging.

web-client/src/app/views/register/register.page.ts
```diff
@@ -1,31 +1,68 @@
...
-    })
-  };
-
+      firstName: ['', Validators.compose([Validators.required])],
+      lastName: ['', Validators.compose([Validators.required])],
+      mobile: [
+        '',
+        Validators.compose([
+          Validators.required,
+          Validators.minLength(1),
```
jongbonga, line 27: I think minLength here should be 10

longtomjr: Min length is 4 and max is 10 from what we discussed.

longtomjr:
Did a rebase to revert the package lock change.

longtomjr (approved):

jongbonga (approved):

PR #55, opened 2021-08-19
feat(web-client): add lock screen

jongbonga:


longtomjr (changes requested):
Not sure if using a custom keypad in this context makes sense. We should be able to mark the input field as only for digits, and a mobile device will pick that up and display a keypad with only digits afaik.

longtomjr (approved):
Looks good!

Issue #27, opened 2021-07-19
Transaction functionality


longtomjr:
## Web Client

 - [x] Sender signs into their wallet using their PIN
 - [x] Receiver provides their Algorand address (As text or QR code) and inputs it into the app
 - Sealed API Request:
   - [x] App sends the Sender's Wallet ID and pin, alongside the Receiver's Address
   - [x] API returns Outcome result (Success/Failure with reasons?)

PRs:
- [x]  #59

## Web Server
 - Sealed enclave request: 
   - [x] Request: Transaction request, with Sending Wallet ID, Sending wallet PIN,
     and recieving wallet address
   - Enclave operation:
     - [x] Verify sending wallet ID + PIN and get the Wallet's private key
     - [x] Create offline transaction (using algo libary) using the sender and reciever's details
   - [x] Return the created transaction

Deferred to client:

 - [x] Submit and verify transaction to Algorand node

PRs:

- [x] #68

PiDelport:
Wrapping this up for now: we can split out any trailing tasks.

Issue #26, opened 2021-07-19
Open wallet functionality


longtomjr:
## Web Client

 - [x] UI: Wallet owner enters Wallet ID (or scan QR code) and PIN
 - Sealed API exchange:
   - [x] Request: Wallet open request, with access credentials
   - Response:
     - [x] Success: Nautilus Wallet ID, owner name, any other wallet details
     - [x] Failure: Bad PIN / Wallet ID
 - [x] UI: Show open wallet view, with owner details and balance

PRs:
- [x] #59

## Web Server
 - Sealed enclave exchange:
   - [x] Request: Wallet open request with access credentials
   - [x] Enclave Operation: Verify the access credentials, and return wallet details
   - Response:
     - [x] Success: Nautilus Wallet ID, owner name, any other wallet details
     - [x] Failure: Bad PIN / Wallet ID

Deferred to client:

- [x] Obtain account balance from Algorand node
 
PRs:

- [x] #68

PiDelport:
Wrapping this up for now: we can split out any trailing tasks.

Issue #25, opened 2021-07-19
Create wallet functionality


longtomjr:
## Web Client

 - Update "Create Wallet" view
   - [x] Add wallet creation form requesting name and pin - confirm pin.
   - [x] Send the details using a sealed api exchange - Depends on: #1
   - [x] Receive the sealed exchange result that contains the new Wallet ID (Algorand address)
   - [x] Prompt the user to save their wallet ID (provide QR-Code that can be printed)
   - [x] Redirect the user to the landing page.

PRs:
- [x] #59
- [x] #39 
- [x] #54 

## Web Server
 - [x] Add ~`/wallet` put~ `/wallet-operation` POST endpoint that takes a sealed payload for wallet creation and passes it to the enclave
 - [x] Unseal the payload inside of the enclave using the sealed api exchange functionality
 - [x] Add wallet creation functionality inside of the enclave using the ported Algonaut functionality
 - [x] Persist algorand account details alongside the pin using the KvStore - Depends on: #4
 - [x] Create a sealed api result containing the wallet ID (algorand address)

PRs:

- [x] #68

PiDelport:
Wrapping this up for now: we can split out any trailing tasks.

PR #59, opened 2021-08-24
feat(web-client): add API schema and service

PiDelport:
### Tasks

- [x] Sort out the `@typescript-eslint/naming-convention` lint situation
- [x] environment types

Issues:

- #25
- #26
- #27

### Deferred tasks

- ~Use https://github.com/capacitor-community/http for http calls~ – the library does not support posting binary data, it seems: we'll need upstream patches.
- WalletService test coverage: split to #74
- Schema test coverage: split to #75

### Follows

- #66 (for Chromatic baseline fix)
- #65 (for `package-lock.json` sync)

### DIFF
--- a/web-client/.eslintrc.json
+++ b/web-client/.eslintrc.json
@@ -36,6 +36,25 @@
             "prefix": "app",
             "style": "camelCase"
           }
+        ],
+        "@typescript-eslint/naming-convention": [
+          "error",
+          {
+            "selector": "default",
+            "format": ["camelCase", "snake_case"]
+          },
+          {
+            "selector": "variable",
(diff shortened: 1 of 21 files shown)


longtomjr (approved):
Looks good thanks. There might be refactoring we can do to make it nicer to work with the union types, but we can do that as a follow-up.

PiDelport:
@longtomjr: I ended up adding some more test coverage for the wallet API operations, and found and fixed some small niggles in the process. The coverage diff is looking a lot better, even though it's not quite on target yet.

Here's the diff since your last review, for reference: [9f80139..cf3f415](https://github.com/registreerocks/nautilus-wallet/compare/9f80139..cf3f415)

The main uncovered code that still remains is the Algorand client code.

PR #64, opened 2021-08-30
payment-confirmation

jongbonga:


web-client/src/app/views/pay/pay.page.ts
```diff
@@ -25,9 +31,75 @@ export class PayPage implements OnInit {
...
+      // if (success) {
+      //   this.notifySuccess(
+      //     'R250',
+      //     'G6AIRDAJFSBXNFBHLQ2F5JLZJ6EYYYLDZSCDHUQUB2YUG5QO4ZB4VNAL7I',
+      //     'LNJH2Q2BYMV7KMMFMEEUCKIW6C3HPBOL4ARZBPPE2OBVVDPNVN6A',
+      //     new Date()
+      //   );
+      // } else {
+      //   this.notifyError(errorMessage);
+      // }
```
PiDelport, line 46: Should we remove or replace this before merging?

longtomjr: This gets touched in the demo branch iirc, so I think it is fine to keep this in for now since we are going to be cleaning up on the demo branch. This is clearly marked as TODO so I think it is fine for now.

PR #73, opened 2021-08-31
test(web-client, LockscreenPage): default autofocus to false, for Chromatic

PiDelport:
This exposes LockscreenPage's `autofocus` as an input property, preserving the default `true`, but disabling it for Storybook.

This should avoid the inconsistent snapshots in Chromatic.

### DIFF
--- a/web-client/src/app/views/lockscreen/lockscreen.page.html
+++ b/web-client/src/app/views/lockscreen/lockscreen.page.html
@@ -19,7 +19,7 @@
             type="password"
             formControlName="code"
             inputmode="tel"
-            autofocus="true"
+            [autofocus]="autofocus"
             minlength="4"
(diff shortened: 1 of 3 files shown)


longtomjr (approved):

PiDelport:
@longtomjr: Okay, even a 5s delay was not enough to get the autofocus consistent, so I'm guessing there are other sources of non-determinism here. (When I just test locally, I can't actually seem to get the autofocus working anyway, so I'm guessing there are complicated browser, state, and/or initialization order factors here.)

It occurred to me that it's simpler and more reliable to just parameterise `autofocus`, and disable it by default for Storybook: I implemented that instead. (That technically invalidates the branch name, but keep it in this PR for expediency.)

PR #78, opened 2021-09-02
notifications

jongbonga:


web-client/src/app/utils/errors/global-error-handler.ts
```diff
@@ -0,0 +1,26 @@
...
+    private navCtrl: NavController
+  ) {}
+
+  handleError(error: Error) {
+    this.zone.run(
+      async () =>
+        await this.notification.swal
+          .fire({
+            icon: 'error',
+            text: error.message || 'Undefined client error',
```
longtomjr, line 19: For errors I would prefer if we don't display the error message for the demo. Just say "Unexpected Error Occurred". Or something like that.

web-client/src/app/utils/errors/http-error.interceptor.ts
```diff
@@ -0,0 +1,34 @@
...
+    request: HttpRequest<any>,
+    next: HttpHandler
+  ): Observable<HttpEvent<any>> {
+    return next.handle(request).pipe(
+      catchError((error: HttpErrorResponse) => {
+        console.error('Error from error interceptor', error);
+
+        this.notification.swal.fire({
+          icon: 'error',
+          text: error.message ?? JSON.stringify(error),
```
longtomjr, line 27: Same, here I would prefer if we just have "Unexpected Error Occurred".

longtomjr (approved):

PR #79, opened 2021-09-03
Follow up for registration page. Added fixed validations

Nghondzweni:


web-client/angular.json
```diff
@@ -168,6 +168,7 @@
     }
   },
   "cli": {
+    "analytics": false,
```
jongbonga, line 171: what's the relevance of this line on the registration page?

Nghondzweni: Oh, post installing angular, it prompted me to share analytics with Google Analytics. I selected no. I think that added that line.  I'll just remove it.

PiDelport:
It looks like these changes got subsumed into #54, and we can close this PR?

PR #86, opened 2021-09-03
manual-address

jongbonga:


web-client/src/app/views/manual-address/manual-address.page.html
```diff
@@ -0,0 +1,37 @@
...
+          [(ngModel)]="address"
+        ></ion-textarea>
+
+        <ion-button
+          expand="block"
+          shape="round"
+          type="submit"
+          class="mt-4"
+          (click)="confirm()"
+          [disabled]="!address"
```
longtomjr, line 31: Is is possible to implement validation / disabled for the wallet address here. See this function I used for the wallet access page: https://github.com/registreerocks/nautilus-wallet/blob/75bd6fd8be549242d905d4c7b0aa6492a298f391/web-client/src/app/views/wallet-access/wallet-access.page.ts#L39-L45

Not sure what the cleanest way would be to implement something like this, but some sort of validation here would help I think.

The `isValidAddress` is from the algosdk. See the imports.

longtomjr (commented):
One minor thing we probably want to get in before the demo

longtomjr (approved):

Issue #96, opened 2021-09-14
CBDC services model


PiDelport:
### To do

See [Technical Architecture](https://github.com/registreerocks/nautilus-wallet/wiki/Technical-Architecture)

- [ ] Base rusty-celery code and scaffolding 
- [ ] Account creation worker module
- [ ] KYC service / worker module (Onfido)
- [ ] Messaging service / worker module (Twilio)

PiDelport:
- Split out: #123

PR #97, opened 2021-09-14
feat(asset-services): base layout for Celery, worker, and API crates

PiDelport:
- Issue: #96


asset-services/asset-services-api/src/main.rs
```diff
@@ -0,0 +1,13 @@
+use std::error::Error;
+
+use asset_services_celery::run_helpers::{build_app, init_logging};
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
```
longtomjr, line 6: How will we call the asset services from the outside. Is the plan to expose http endpoints for that? If so we probably just need to chat about if we are going to pull in actix-web for this as well or if we want to look at another option.

PiDelport: Yup, the plan is to expose a web API here, and to use whatever's most pragmatic for that.

PiDelport: Update: I added a placeholder "ping" task and endpoint using Axum.

asset-services/asset-services-celery/src/celery_box.rs
```diff
@@ -0,0 +1,106 @@
...
+
+/// Wrap [`Celery`] struct.
+pub enum CeleryBox {
+    Amqp(Arc<Celery<AMQPBroker>>),
+    Redis(Arc<Celery<RedisBroker>>),
+}
+
+impl CeleryBox {
+    /// Build an instance using [`Celery::builder`], with the broker determined by the scheme of `broker_url`.
+    pub async fn build(name: &str, broker_url: &str) -> Result<Self, CeleryError> {
```
Nghondzweni, line 25: @PiDelport I see here the broker instance can either be AMQP or Redis. How does it determine which broker instance to initialize? Is this set in the environment variable in docker compose 'BROKER_URL'?

Nghondzweni (approved):

asset-services/asset-services-celery/src/celery_box.rs
PiDelport: Yup, this helper determines it at runtime by looking at the URL scheme of the `broker_url` passed in by `init_celery` / `init_celery_from_env`.

It's not ideal that we need this wrapper, but it makes it possible for us to switch between RabbitMQ and Redis without editing and recompiling all the downstream references every time, which would make testing broker-related things very painful.

This is something that rusty-celery should support better: I plan to file an issue about it.

longtomjr (approved):

PR #106, opened 2021-09-26
feat(asset-services): add RedisTaskRecords persistence for Celery

PiDelport:
- Issue: #96 

- Follows #105


asset-services/asset-services-api/src/errors.rs
```diff
@@ -1,25 +1,28 @@
...
 
-use asset_services_celery::exports::celery_prelude::CeleryError;
 use axum::http::{Response, StatusCode};
 use axum::response::IntoResponse;
 
-/// Map Celery errors to Internal Server Error responses.
-pub struct CeleryErrorResponse(CeleryError);
+/// Represent arbitrary errors as Axum Internal Server Error responses using [`anyhow`].
+#[repr(transparent)]
+pub struct AnyhowErrorResponse(anyhow::Error);
```
longtomjr, line 8: I am not 100% sure what the need is for specifying `repr(transparent)` here. Are we transmuting between the wrapping and inner type, or is this to get the `.context()` methods to work with the wrapping type?

longtomjr (approved):

Nghondzweni (approved):

asset-services/asset-services-api/src/errors.rs
PiDelport: It's not required, but I like specifying it as a form of executable documentation that the struct is intended as a newtype wrapper, to remove ambiguity (analogous to Haskell's distinction between the `newtype` and `data` keywords).

The main practical upshot is that if anything changes that stops it from being a newtype, then this will raise a compiler error for investigation rather than silently proceeding.

PiDelport:
@longtomjr, @Nghondzweni: I pushed a round of cleanups, and finished implementing task result returning for the verification handlers too:

* [f15e58d..c8819aa](https://github.com/registreerocks/nautilus-wallet/compare/f15e58d..c8819aa)

PR #107, opened 2021-09-27
feat(asset-services,onfido): add Onfido client

PiDelport:
- Issue: #96 

- Follows #106


asset-services/onfido/src/client.rs
```diff
@@ -0,0 +1,119 @@
...
+pub struct OnfidoClient {
+    configuration: Configuration,
+}
+
+impl OnfidoClient {
+    fn new(configuration: Configuration) -> Self {
+        Self { configuration }
+    }
+
+    pub fn from_env() -> Result<OnfidoClient, env_vars::EnvVarError> {
```
longtomjr, line 18: Minor thing here. I am not 100% sure with the Onfido client getting the env vars for configuration. I would prefer if the new method were made public, and the env vars would be retrieved and passed in when instantiating the client. This makes the onfido client implementation rely on the environment.

Practically I think the `init_onfido_client` function in kyc.rs should be responsible for getting the env variable, and instantiating the client.

longtomjr (changes requested):
Looks good. Just a minor clarification/change.

Nghondzweni (approved):

asset-services/onfido/src/client.rs
PiDelport: I'd motivate for keeping `from_env` in the library, though. It's not uncommon for SDK-style libraries to allow optionally configure themselves from a set of known environment variables, and my hope is that we can eventually split these clients out to stand-alone libraries like that.

The motivation for letting application code delegate this configuration to the library is to:

1. Reduce the amount of boilerplate code for the common case.

2. Avoid unintentional environment variable name and format differences.

3. Allow optional variables and future extensions without application code having to know about it.

   For example, Twilio's libraries support optional variables like `TWILIO_REGION`  `TWILIO_EDGE`, which can specified in deployment configurations even if the application code didn't anticipate the need to support them: we'll probably want to add support for those to the library code as we need them.
   
   Onfido similarly has regions and other deployment configuration that we'll probably want to support configuring in a common way as we move to production.

longtomjr (approved):

PR #109, opened 2021-09-30
Services update

jongbonga:


web-client/src/app/stores/session/session.service.ts
```diff
@@ -0,0 +1,16 @@
...
+import { SessionStore } from './session.store';
+
+@Injectable({ providedIn: 'root' })
+export class SessionService {
+
+  constructor(private sessionStore: SessionStore, private http: HttpClient) {
+  }
+
+  get() {
+    return this.http.get('').pipe(tap(entities => this.sessionStore.update(entities)));
```
longtomjr, line 13: I am not sure I understand need for an http call here. Is this how it would look if we were using a conventional API?

longtomjr (approved):
Thanks a lot for getting this sorted. Looks good!

web-client/src/app/stores/session/session.service.ts
jongbonga: not needed, removing it.

longtomjr: Ah ok, thanks!

PR #71, opened 2021-08-31
Demo branch part 5

longtomjr:
We should pull out/refactor what is here into separate PRs so that we can have the flow implemented.

We also need to just go through the steps and go over the error handling where needed.

TODO:
- [x] Delay after confirming pin and loading screen displaying
- [x] Add text to the transaction loading screen that tells the user it will take a few seconds to be confirmed

TODO Refactor:
- [x] Refactor the wallet store, query, etc. to be called session store, query etc.
- [x] Merge/Refactor the `NewWalletService` and the `WalletService`. Not sure what separation will make the most sense in this context.

----

PR group:

- #111
- #112
- #113
- #114
- #71

PR #111, opened 2021-10-04
Demo branch part 1

PiDelport:
PR group:

- #111
- #112
- #113
- #114
- #71

### DIFF
--- a/web-client/src/app/services/wallet/wallet.service.ts
+++ b/web-client/src/app/services/wallet/wallet.service.ts
@@ -1,0 +1,36 @@
+import { Injectable } from '@angular/core';
+import { EnclaveService } from 'src/app/services/enclave/enclave.service';
+import { SessionStore } from 'src/app/stores/session/session.store';
+import { WalletDisplay } from 'src/schema/entities';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class WalletService {
+  constructor(
+    private sessionStore: SessionStore,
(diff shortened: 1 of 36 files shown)


longtomjr (approved):

web-client/src/app/services/wallet/wallet.service.ts
```diff
@@ -0,0 +1,81 @@
...
+      }
+    } catch (err) {
+      this.sessionStore.setError(err);
+    }
+  }
+
+  async openWallet(walletId: string, pin: string) {
+    try {
+      const res = await this.enclaveService.openWallet({
+        wallet_id: walletId,
```
jongbonga, line 39: Maybe we should have consistency between the enclave service vars & wallet `(walletId & wallet_id)`

PiDelport: The `snake_case` identifiers are what the MessagePack protocol uses: the arguments here are directly part of that.

We could add another layer with translation code that converts between the `snake_case` and `camelCase` conventions, but this could get complex: we'd either need to implement something that does dynamic conversions (which would be hard to get type-checked), or we'll need something that can automatically generate the right variants and conversions. (Or we could manually write and maintain them, but I wouldn't be keen for that, from a maintenance perspective.)

Using the protocol identifiers directly means there's just one thing to keep in sync for now, so pragmatically, I'm erring towards that.

jongbonga (approved):

PR #124, opened 2021-10-07
feat(asset-services): add Celery worker preflight checks

PiDelport:
- Issue: #96

This checks various configuration and operational pre-conditions for the Celery workers and tasks, before accepting tasks from the queue.

### DIFF
--- a/asset-services/docker-compose.yaml
+++ b/asset-services/docker-compose.yaml
@@ -31,6 +31,8 @@ services:
     init: true
     environment:
       BROKER_URL: "redis://redis:6379"
+      PREFLIGHT_CHECK_LEVEL:
+
       ONFIDO_API_TOKEN:
       TWILIO_ACCOUNT_SID:
       TWILIO_AUTH_TOKEN:


asset-services/asset-services-celery/src/preflight/checks.rs
```diff
@@ -0,0 +1,52 @@
...
+use crate::preflight::types::CheckLevel;
+use crate::task_records::redis_impl::RedisTaskRecords;
+use crate::tasks;
+
+const LOCAL: &str = "preflight::local";
+const REMOTE: &str = "preflight::remote";
+
+/// Check [`RedisTaskRecords`] connectivity.
+///
+/// This doubles
```
longtomjr, line 12: I think some text in the comment is missing here

asset-services/asset-services-celery/src/preflight/types.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use env_var_helpers::env_vars;
+use strum::EnumString;
+
+/// Specify what level of preflight checks to run (inclusive).
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Debug)] // core
+#[derive(EnumString)] // strum
+#[strum(ascii_case_insensitive)]
+#[strum(serialize_all = "snake_case")]
+pub enum CheckLevel {
```
longtomjr, line 13: This representation does not seem super clear from just looking at the code and types.

Three options I see that might make it clearer:
 1. Since we are comparing them similar to numeric values in the `impl`, maybe assign a spesific numberic value to the enum values
 2. Use a bitflags style enum (probably overkill for this)
 3. Name the values `None, Local, All` instead. - (I think I prefer this)


This is not a major issue, since the comments do clarify the intent, just came up during review.

asset-services/twilio-verify/src/client.rs
```diff
@@ -29,6 +29,31 @@ impl VerifyClient {
         ))
     }
 
+    /// Fetch the verification service configuration for the current service.
+    pub async fn fetch_current_service(
```
longtomjr, line 33: Not a 100% on the twillio api conventions, but it might be a little clearer to call this `fetch_current_service_config`. (same for the `fetch_service` below).

Unfortunately "Service" has a lot of meanings, so a little bit more clarity might help.

PiDelport: Yeah, the Twilio API is remarkably Not Well Named here. 😄

The Twilio Verify API docs use "Service" (or "Verification Service") consistently, so I erred towards matching that terminology and convention here: I figure it's better to have to learn their resource naming convention once, rather than having to deal with both their naming convention and a slightly different one for our client.

PR #136, opened 2021-11-22
refactor(web-client): extract PayFromToComponent

jongbonga:


web-client/src/app/views/pay/pay.module.ts
```diff
@@ -17,6 +18,6 @@ import { PayPage } from './pay.page';
     SharedModule,
     InputMaskModule.forRoot({ inputSelector: 'input', isAsync: true }),
   ],
-  declarations: [PayPage],
+  declarations: [PayPage, PayFromToComponent],
```
PiDelport, line 21: It might make more sense to have a shared module under `src/app/components` to declare the components, rather than having the components under `src/app/components` but their declarations under `src/app/views`?

jongbonga: 💯

web-client/src/app/views/pay/pay.page.stories.ts
```diff
@@ -29,8 +31,3 @@ const Template: Story<Args> = ({ receiverAddress, ...state }: Args) => ({
 });
 
 export const Default = Template.bind({});
-Default.args = {
-  name: 'Test Owner',
-  balance: 1234,
-  receiverAddress: 'G6AIRDAJFSBXNFBHLQ2F5JLZJ6EYYYLDZSCDHUQUB2YUG5QO4ZB4VNAL7I',
-};
```
PiDelport, line 36: Can we re-use the extract component story here, so that the story still renders with the placeholders like before? (See the [Chromatic change / comment](https://www.chromatic.com/test?appId=619e482ab3e38b003a063523&id=619e4abeae73be003a7189ba))

The Storybook docs cover some of the ways to approach this: https://storybook.js.org/docs/angular/writing-stories/stories-for-multiple-components

web-client/src/app/views/pay/pay.page.stories.ts
```diff
@@ -21,6 +22,7 @@ type Args = { receiverAddress: string } & SessionState;
 
 const Template: Story<Args> = ({ receiverAddress, ...state }: Args) => ({
   moduleMetadata: {
+    declarations: [PayFromToComponent],
```
PiDelport, line 25: This declaration should be removed: it conflicts with the one in `SharedModule`.

(This doesn't actually present in Chromatic, but it makes the story fail when running Storybook locally.)

PiDelport (changes requested):

PiDelport (approved):
I added some changes which should resolve the open feedback.

web-client/src/app/views/pay/pay.page.stories.ts
PiDelport: e63e060 should resolve this.

web-client/src/app/components/pay-from-to/pay-from-to.component.ts
```diff
@@ -0,0 +1,16 @@
+import { Component, Input, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'app-pay-from-to',
+  templateUrl: './pay-from-to.component.html',
+  styleUrls: ['./pay-from-to.component.scss'],
```
billguo99, line 6: is `styleUrls` compulsory? The `pay-from-to.component.scss` file is empty.

billguo99 (approved):
My comments are not blocking, just 1 question and a space missing from a line. Looks good :)

web-client/src/app/components/pay-from-to/pay-from-to.component.ts
jongbonga: Every component/page get a style file upon creation.
It doesn't really matter if it's empty or not because it's dealt with on build with the [tree shaking](https://developer.mozilla.org/en-US/docs/Glossary/Tree_shaking)

web-client/src/app/views/pay/pay.page.html
```diff
@@ -9,30 +9,11 @@ <h1 class="font-semibold font-nasalization">Pay</h1>
...
-              <ion-label class="ion-text-wrap">
-                <h2 class="!font-black">To:</h2>
-                <p>{{wallet}}</p>
-              </ion-label>
-            </ion-item>
-          </ion-list>
-        </ion-card>
+        <app-pay-from-to
+          [name]="sessionQuery.name | async"
+          [balance]="sessionQuery.balance| async"
```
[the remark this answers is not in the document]
jongbonga: Prettier should've dealt with it normally. Thanks

PR #139, opened 2021-12-02
feat(web-client): add sentry

jongbonga:


longtomjr:
For the error handling, we already have this global error handler that we are using: https://github.com/ntls-io/nautilus-wallet/blob/main/web-client/src/app/utils/errors/global-error-handler.ts

We might need to do a slightly more complex integration with Sentry that takes into account the global handler we already have.

It is probably worth looking at the global error handling currently, and see if it make sense. Some of it might not be 100% needed or correct if we integrate Sentry now.

jongbonga:
> For the error handling, we already have this global error handler that we are using: https://github.com/ntls-io/nautilus-wallet/blob/main/web-client/src/app/utils/errors/global-error-handler.ts
> 
> We might need to do a slightly more complex integration with Sentry that takes into account the global handler we already have.
> 
> It is probably worth looking at the global error handling currently, and see if it make sense. Some of it might not be 100% needed or correct if we integrate Sentry now.

The global error will reset the store and log the user out as it's currently doing, but Sentry will be able to trace the source with this current basic configuration (at least).

longtomjr:
> > For the error handling, we already have this global error handler that we are using: https://github.com/ntls-io/nautilus-wallet/blob/main/web-client/src/app/utils/errors/global-error-handler.ts
> > We might need to do a slightly more complex integration with Sentry that takes into account the global handler we already have.
> > It is probably worth looking at the global error handling currently, and see if it make sense. Some of it might not be 100% needed or correct if we integrate Sentry now.
> 
> The global error will reset the store and log the user out as it's currently doing, but Sentry will be able to trace the source with this current basic configuration (at least).

Do we know the order that this will run in, and is there an explicit way to define that?

jongbonga:
> > > For the error handling, we already have this global error handler that we are using: https://github.com/ntls-io/nautilus-wallet/blob/main/web-client/src/app/utils/errors/global-error-handler.ts
> > > We might need to do a slightly more complex integration with Sentry that takes into account the global handler we already have.
> > > It is probably worth looking at the global error handling currently, and see if it make sense. Some of it might not be 100% needed or correct if we integrate Sentry now.
> > 
> > 
> > The global error will reset the store and log the user out as it's currently doing, but Sentry will be able to trace the source with this current basic configuration (at least).
> 
> Do we know the order that this will run in, and is there an explicit way to define that?

They both run on app init but serve 2 different purposes, sentry will record and the global will trigger the rest

longtomjr (approved):
Looks good. Thanks

PR #144, opened 2021-12-03
fix(web-client): fix lockscreen submission

### LINKED ISSUE #141: bug: Pin modal issue
### Current Behavior

The modal for pin validation submit the value when the modal is dismissed

### Expected Behavior
The modal should only submit the value when validated. when modal is dismissed, it should behave accordingly and do nothing.

### Steps to Reproduce
When modal for pin open, enter a value then dismiss the modal

### Additional Information

jongbonga:
fix issue where pin will be checked on lockscreen dismiss

Closes #141


web-client/src/app/views/wallet-access/wallet-access.page.ts
```diff
@@ -46,26 +42,25 @@ export class WalletAccessPage implements OnInit {
     const address = value?.trim();
 
     if (address && isValidAddress(address)) {
-      const pinPromise = this.presentLock();
+      const pinPromise = await this.presentLock();
```
longtomjr, line 45: Double check that the loading control gets shown correctly on mobile. I ran into an issue where there is downtime between canceling or entering the pin, and when the loading indicator displays, so that is why the promise gets awaited later.

jongbonga: I will do that and report back

jongbonga: @longtomjr I haven't experienced any issue with the loading indicator

longtomjr (approved):

PR #146, opened 2022-01-06
build(web-client): upgrade to angular 13

jongbonga:


web-client/src/app/components/action-item/action-item.component.html
```diff
@@ -24,7 +24,7 @@
     detailIcon="chevron-forward-circle"
     class="app-item"
     [disabled]="disabled"
-    [routerLink]="path"
+    [routerLink]="path ?? []"
```
billguo99, line 27: What does adding `?? []` do?

web-client/tsconfig.json
```diff
@@ -24,6 +24,7 @@
     "enableI18nLegacyMessageIdFormat": false,
     "strictInjectionParameters": true,
     "strictInputAccessModifiers": true,
-    "strictTemplates": true
+    "strictTemplates": true,
+    "allowSyntheticDefaultImports": true
```
billguo99, line 28: I see the use of setting this to `true` - https://www.typescriptlang.org/tsconfig#allowSyntheticDefaultImports but I'm not sure why we need it.

billguo99 (commented):
Running `npm install`:
```
57 vulnerabilities (2 low, 20 moderate, 34 high, 1 critical)
```

Can run `npm audit fix` can fix a couple of vulnerabilities without any breaking changes:
```
46 vulnerabilities (2 low, 15 moderate, 29 high)
```

web-client/tsconfig.json
jongbonga: This property was added by the Angular CLI. 
it should cover the cases whereby we have things import like [this](https://github.com/ntls-io/nautilus-wallet/blob/5916aea15fd3a6920931a56e3a33667532189750/web-client/src/tests/transloco.helpers.ts)

billguo99 (dismissed):

billguo99 (approved):
The ScannerPage test works locally, but fails intermittently in Github actions. As discussed in retro, we can approve and merge.

longtomjr (approved):
Looks good. Thanks a lot for getting this sorted.

PiDelport (commented):
- Follow-up: #160

The upgrade seems have broken the use of `templateUrl` in the theme story: I'm not sure why, yet, but using an inline `template` instead seems to work, so I did that as a workaround for the time being.

PR #132, opened 2021-10-29
feat(web-client): support browser autocomplete for wallet address / PIN

PiDelport:
This also declares autocomplete for the registration form's name and mobile number fields.

This is mainly to ease repeated sign-ins and registrations while developing, which can get very laborious otherwise.

- Depends on https://github.com/ntls-io/nautilus-wallet/pull/148

On hold until we have the feature flags to conditionally enable this for development.

PR #148, opened 2022-01-07
build(web-client): add .node-version file, for Netlify

PiDelport:
Docs:

* https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-and-javascript
* https://github.com/shadowspawn/node-version-usage

Netlify was previous building with Node 12 and NPM 6 (lockfile v1), which is definitely much older than we're comfortable with.

Blocks:
- #132
- #149


PiDelport:
This might fix the build failure in #132! 🎉

PR #157, opened 2022-02-24
docs(book): initial version

PiDelport:
This ports some of our technical documentation to a more readable mdbook version.


PiDelport:
Published with [gh-pages](https://github.com/ntls-io/nautilus-wallet/tree/gh-pages) here: https://ntls-io.github.io/nautilus-wallet/

jeanpierredevilliers:
The included images seem too detailed and I found them hard to read.  I suspect they may better suited for architectural purposes than for developer documentation.  What are your thoughts on this?

docs/book/src/design/overview.md
```diff
@@ -0,0 +1,21 @@
...
+
+The Nautilus Wallet system consists of the following high systems:
+
+1. An underlying [ledger platform](ledger-platform.md) and its asset(s)
+2. The [Wallet](wallet.md) itself:
+   - The [Wallet TEE](wallet-tee.md) (server) hosts and provides access to the Wallet SGX enclave
+   - The [Wallet Connector app](wallet-connector-app.md) (client) securely connects wallet owners with their wallets
+3. [Asset services](asset-services.md), which help manage and govern the asset:
+   - The [KYC service](kyc-service.md)
+   - User verification service
```
jeanpierredevilliers, line 11: - We might, at the end of the book, want to add a glossary of terms.  We could even link it here to draw attention to it early on.  The topic at hand is very dense when it comes to the jargon involved.
- The `kyc-service.md` link appears to be broken.  Forgotten commit or intentional?

docs/book/src/design/overview.md
```diff
@@ -0,0 +1,21 @@
...
+
+## System Landscape diagram
+
+[![System Landscape diagram][diagram]][diagram]
+
+[diagram]: diagrams/SystemLandscape.svg
```
jeanpierredevilliers, line 17: See [my comment](https://github.com/ntls-io/nautilus-wallet/pull/157#issuecomment-1049923365) regarding the imported images in general

docs/book/src/design/ledger-platform.md
```diff
@@ -0,0 +1,32 @@
...
+## Algorand
+
+The [Algorand] ledger supports transacting using [Algorand Standard Assets (ASAs)][ASA].
+
+The Wallet TEE uses an SGX port of the [algonaut] SDK to implement key generation and transaction signing.
+
+The Wallet Connector uses [js-algorand-sdk] to communicate with the Algorand network.
+
+[Algorand]: https://www.algorand.com/
+[ASA]: (https://developer.algorand.org/docs/get-details/asa/)
```
jeanpierredevilliers, line 16: Link broken.  Perhaps it should be the following:
```suggestion
[ASA]: https://developer.algorand.org/docs/get-details/asa/
```

docs/book/src/design/wallet.md
```diff
@@ -0,0 +1,13 @@
+# Wallet
+
+This provides the functionality of a non-custodial wallet to end-users without smartphones, accessible through a smartphone-based connector app with limited trust.
```
jeanpierredevilliers, line 3: This requires some elaboration.  It is not immediately clear how an application can be both "smartphone-based" while also targeting users *without* smartphones.  I assume this is where the "limited trust" comes in but there is value in being explicit here.

docs/book/src/design/wallet-tee.md
```diff
@@ -0,0 +1,25 @@
+# Wallet TEE
+
+The Wallet TEE consists of a trusted SGX enclave, alongside an untrusted host service.
```
jeanpierredevilliers, line 3: Perhaps link an overview of SGX, like so
```suggestion
 The Wallet TEE consists of a trusted [Intel SGX enclave](https://www.intel.com/content/www/us/en/developer/tools/software-guard-extensions/overview.html), alongside an untrusted host service.
```

docs/book/src/design/wallet-tee.md
```diff
@@ -0,0 +1,25 @@
...
+
+## SGX enclave
+
+The Wallet SGX enclave contains the system's [Trusted Computing Base (TCB)][TCB], and is responsible for all security-sensitive wallet operations:
+
+- Generating and managing account keypairs
+- Performing privileged operations such as transaction signing
+
+Once generated, secret keys and other sensitive data never leaves the enclave unencrypted. All persistent storage is cryptographically sealed and authenticated, and all trusted communication with the enclave occurs over cryptographically attested and mutually-authenticated channels.
```
jeanpierredevilliers, line 12: As it is specific to the context of TEEs, the concept of *sealed storage* may need defining

docs/book/src/design/wallet-tee.md
```diff
@@ -0,0 +1,25 @@
...
+
+[TCB]: https://en.wikipedia.org/wiki/Trusted_computing_base
+
+## Host service
+
+The SGX enclave is embedded in an untrusted host service, which facilitates communication between the enclave and the outside world.
+
+* Remote attestation, which allows the Connector app to remotely verify the identity and integrity of the enclave, and exchange keys to set up a trusted communication channel
```
jeanpierredevilliers, line 20: At this point in the documentation, it has not yet been made clear what the *Connector app* is.  I would suggest one of the following:

1. Change the order of the chapters so that the "Connector app" chapter appears before this one
2. Keep the order the same but refer to the Connector simply as the "client application"

docs/book/src/design/wallet-connector-app.md
```diff
@@ -0,0 +1,7 @@
+# Wallet Connector app
+
+The Wallet Connector app runs on web and smartphone (Android, iOS) devices, and allows users other than the device owner to create and access their wallets, using a combination of QR code and PIN based authentication.
```
jeanpierredevilliers, line 3: - Somewhere, it should probably be made explicit that the *Connector app* is the client application
- This paragraph seems to suggest that parties other than the wallet owner has access to the funds of the latter

docs/book/src/design/asset-services.md
```diff
@@ -0,0 +1,15 @@
...
+
+[diagram]: diagrams/CBDCServices-Container.svg
+
+**Figure:** [Container diagram] of the Nautilus asset services.
+
+[Container diagram]: https://c4model.com/#ContainerDiagram
+
+## Source code
+
+* <https://github.com/ntls-io/nautilus-wallet/tree/main/asset-services>
```
jeanpierredevilliers, line 15: For those less familiar with the ecosystem surrounding *digital assets*, I would suggest an  explanation or definition as well as some concrete example.  Similarly for the notion of an *asset service*.

jeanpierredevilliers (commented):
Overall, the document paints a good picture.  Some care, however, might be needed for audience members less familiar with TEEs, digital assets and so forth.

docs/book/src/design/overview.md
PiDelport: A glossary is definitely a good idea.

The KYC service page still needs to be created, yeah: good catch! I removed the link, for now.

docs/book/src/design/ledger-platform.md
PiDelport: Good catch, thanks!

docs/book/src/design/wallet-tee.md
PiDelport: We do link to this in the main overview.

My instinct would be to assume some general familiarity with Intel SGX once it gets into this level of detail (documenting individual system components), and to limit the links to general SGX introductory material more to the opening page, glossary, or a dedicated SGX / TEE background section?

docs/book/src/design/wallet-tee.md
PiDelport: Yup, this can probably go in the glossary alongside other SGX concepts.

docs/book/src/design/wallet-tee.md
PiDelport: Yeah; we should probably give more context for this in the parent Wallet chapter.

I don't think changing the order addresses the dependency problem: describing the Connector app client component also doesn't make sense without describing the TEE server component, so either order will require a sibling reference. (I weakly lean towards keeping the TEE server first, just because it's more architecturally foundational.)

For the naming, I lean towards "Connector app" mainly for consistency with our other documentation and material. It may not be the final name, but we should probably stick to it wherever we can, for now, if only so that it's easier to find and update it later.

docs/book/src/design/asset-services.md
PiDelport: Yup, this still needs to have an overview filled in.

docs/book/src/design/wallet-connector-app.md
PiDelport: Good catch. How about this?

```suggestion
The Wallet Connector app runs on web and smartphone (Android, iOS) devices, and allows transacting users to manage wallets and authorise transactions using a combination of QR codes and PIN based authentication.

The Connector app is primarily intended to facilitate transactions for transacting users other than the device owner (such as customers, friends, or family members, friends, or customers),
so each session that the Connector app establishes with the Wallet TEE is designed to be interactive, private, and short-lived. 
The Connector app retains no session state between sessions, and device owners and different transacting users have no access to each others' wallets or accounts.
```

docs/book/src/design/overview.md
PiDelport: > The included images seem too detailed and I found them hard to read. I suspect they may better suited for architectural purposes than for developer documentation. What are your thoughts on this?

Well, these docs should be covering the system's architecture, so that's probably in scope, at least?

Regarding the level of detail, we may want to have both simplified and more detailed versions. Structurizr is able to toggle descriptions in its detailed view: I'll see if I can wrangle something with that. (We'll probably want to preserve the more detailed versions somewhere: Co has indicated that that's useful.)

docs/book/src/design/wallet.md
PiDelport: How about something like this?

```suggestion
This provides the functionality of a non-custodial wallet to end-users who don't own a smartphone themselves, accessible through a smartphone-based connector app with limited trust.
```

jeanpierredevilliers: I think that is much clearer yes!

docs/book/src/design/wallet-tee.md
jeanpierredevilliers: Ah yes, I concur

docs/book/src/design/wallet-connector-app.md
jeanpierredevilliers: I think this eliminates any ambiguity yes :+1:

docs/book/src/design/overview.md
jeanpierredevilliers: I agree, there is value in the more detailed version.  I only think that all that text is probably a bit excessive for the developer docs.  For this specific use case it can probably be trimmed down a bit, like you mention.  Otherwise, is there a way for it to be togglable?

docs/book/src/design/wallet-tee.md
jeanpierredevilliers: Oh, I missed that one.  Agreed, it is sufficient to link only the first occurence and to assume it understood thereafter.

docs/book/src/design/wallet-tee.md
jeanpierredevilliers: I agree, this can de done in a simple fashion in the parent chapter.  I did not consider that possibiliity but it is certainly much more elegant! :+1:

docs/book/src/design/asset-services.md
jeanpierredevilliers: Understood.  I was only trying to be thorough :nerd_face:

longtomjr (approved):
I scanned through this and I am happy getting this merged in once all the feedback has been sorted (or pulled out into a separate follow up issue). Thanks a lot!

docs/book/src/design/wallet-tee.md
PiDelport: Should we change anything in the parent chapter to resolve this, or is it okay as-is, for now?

docs/book/src/design/overview.md
PiDelport: The dynamic toggling capability won't be trivial to replicate outside of Structurizr UI, I think.

jeanpierredevilliers: I see, ok, no toggling then.  May I suggest then that we display the simplified version but have a link in the caption (or elsewhere) to the detailed versions?  Unless you know of a better solution that is.

docs/book/src/design/wallet-tee.md
jeanpierredevilliers: I think as it stands, it is sufficiently minor so as to address in a future PR.  I'm pulling it out into a new issue for bookkeeping.

docs/book/src/design/asset-services.md
jeanpierredevilliers: This can be addressed in a follow-up PR and it's not like we'll forget about it (in this case) so this can be resolved

jeanpierredevilliers (approved):
I think this a successful initial version and it is ready to be merged.  I've pulled the remaining (mostly minor) loose ends into issue #158

PiDelport:
Cool, I added GitHub Actions workflows to build and deploy the book. I left the build to run for every push, but made the deploy to `gh-pages` a manual action: I figure that's the best balance for now.

PR #162, opened 2022-03-18
Toggle between currencies

jongbonga:


PiDelport:
For the size of this PR, and the number of components added, I think it would make sense to rebase it into a set of smaller PRs for easier review?

PR #131, opened 2021-10-28
feat(web-client): Algorand ASA support

PiDelport:
- Issue: #92

### Related PRs

- Follows https://github.com/ntls-io/nautilus-wallet/pull/173
- Precedes https://github.com/ntls-io/nautilus-wallet/pull/122

PR #166, opened 2022-04-14
refactor(web-client): UI and Storybook improvements

PiDelport:
This accumulates some changes from the following PR:

- #131


web-client/tailwind.config.js
```diff
@@ -23,32 +22,6 @@ module.exports = {
       },
     },
   },
-  variants: {
```
IscoRuta98, line 26: @PiDelport, now that tailwind css v3.0 automatically enables all variants. how would we go about disabling variants that are not needed for the wallet (unless the wallet makes use of all the variants)?

PiDelport: I think with Tailwind v3.0, the JIT engine by default takes care of only generating the variants that are actually used, so we don't need to do anything special for this:

- <https://tailwindcss.com/blog/tailwindcss-v3#just-in-time-all-the-time>

web-client/src/app/components/profile-card-horizontal/profile-card-horizontal.component.ts
```diff
@@ -6,9 +6,9 @@ import { Component, Input, OnInit } from '@angular/core';
   styleUrls: ['./profile-card-horizontal.component.scss'],
 })
 export class ProfileCardHorizontalComponent implements OnInit {
-  @Input() title: string | undefined;
-  @Input() fullName!: string;
-  @Input() cellphone!: string;
+  @Input() title?: string;
+  @Input() fullName?: string;
```
longtomjr, line 10: Did you intend to change the `!` to `?` here as well?

longtomjr (approved):
Looks good. This cleans up a lot of stuff!

Just one minor thing I picked up.

web-client/src/app/components/profile-card-horizontal/profile-card-horizontal.component.ts
PiDelport: Yup: I'm generally in favour of using `?` and checked accessors rather than `!`, to have more predictable crashes and error reporting if the condition ever gets violated..

PR #167, opened 2022-04-17
refactor(web-client): PIN entry

PiDelport:
Summary:

- Factor out a pure `PinEntryComponent`
- Replace `LockscreenPage` with a simpler inline modal in `WalletAccessPage`
- Add some form validation helper code

### Related PRs

- Follows https://github.com/ntls-io/nautilus-wallet/pull/166
- Precedes https://github.com/ntls-io/nautilus-wallet/pull/168

PR #168, opened 2022-04-19
feat(web-client): balance summary & refresh button

PiDelport:
This adds a basic balance summary and refresh button to the wallet page.

### Related PRs

- Follows https://github.com/ntls-io/nautilus-wallet/pull/167
- Precedes https://github.com/ntls-io/nautilus-wallet/pull/169


longtomjr (approved):
Not a fan of the font used for the currencies, but this is something we can look at, at a later date.

PiDelport:
@longtomjr: You're right: that's something that affects the follow-up branches too, so I went ahead and just made it the bold default font instead.

PR #171, opened 2022-04-19
test(web-client): add karma-mocha-reporter

PiDelport:
This makes running the tests much more useful.

See: https://github.com/litixsoft/karma-mocha-reporter#readme


IscoRuta98 (commented):
@PiDelport, the changes look ready to go. 
Perhaps a suggestion: the [karma-mocha-reporter readme](https://github.com/litixsoft/karma-mocha-reporter#readme) explains that there is a showDiff option that can be enabled. Would it not make debugging test failures easier if showDiff is set to true (in the karma.conf.js file) so that we know what output is expected, and what the code actually gives as output?

longtomjr (approved):

PiDelport:
@IscoRuta98: The `showDiff` output looks nice, yeah, but trying it out, it requires some additional dependencies:

```
Error loading module mocha!
You have enabled diff output. That only works with karma-mocha and mocha installed!
Run the following command in your command line:
  npm install karma-mocha mocha diff
```

I'm a bit hesitant to mess with the stock test setup too much right now, in case it causes trouble, so I think we can probably look at getting that working later?

PR #174, opened 2022-04-24
refactor(web-client): remove HttpErrorInterceptor

PiDelport:
GlobalErrorHandler already implements a fallback dialog.

Having this interceptor makes it impossible for application code to handle expected unsuccessful API calls.


IscoRuta98 (approved):
Everything looks good here.

PR #169, opened 2022-04-19
refactor(web-client): payment component structure

PiDelport:
This restructures the payment page into a collection of components, with UI state management aimed at supporting multiple accounts / assets.

### Related PRs

- Follows https://github.com/ntls-io/nautilus-wallet/pull/168

Issue #115, opened 2021-10-04
XRPL support


PiDelport:
### Status

This is currently running for the proof-of-concept demo deployment, but on hold for merging into main: we need to restructure some things. —Pi

### Tasks

- [x] Port libraries to SGX: [ripple-address-codec-rust-sgx](https://github.com/registreerocks/ripple-address-codec-rust-sgx) & [ripple-keypairs-rust-sgx](https://github.com/registreerocks/ripple-keypairs-rust-sgx) (@PiDelport)

### PRs

- [x] ~#117~ (subsumed)
- [x] ~#121~ (subsumed)
- [x] https://github.com/ntls-io/nautilus-wallet/pull/176
- [x] #122
- [x] #118
- [x] #120
- [x] Branch: [xrp-mask...xrp-demo](https://github.com/registreerocks/nautilus-wallet/compare/xrp-mask...xrp-demo) (This is what's deployed to <https://xrpl-demo.ntls.io/>)

### Deployment

- [x] Set up demo instance for for `xrpl-api.registree.io` (@PiDelport) - see [nautilus-wallet-demo@xrp-demo](https://github.com/registreerocks/nautilus-wallet-demo/tree/xrp-demo)
- [x] Configure DNS for `xrpl-api.registree.io` (@longtomjr)

PiDelport:
We should be able to consider this complete now.

PR #176, opened 2022-04-25
feat: add enclave XRPL support

PiDelport:
This adds XRPL support to the `web-server` enclave, and corresponding representation types to the `web-client` schema.

Note that this branch currently switches from doubling the Algorand account as the wallet ID to doubling the XRPL account as wallet ID: see `key_from_id`. We should file a follow-up issue to address this.

### Related

- Issue: https://github.com/ntls-io/nautilus-wallet/issues/115
- Follows https://github.com/ntls-io/nautilus-wallet/pull/179
- Precedes https://github.com/ntls-io/nautilus-wallet/pull/169


longtomjr (approved):
Looks good. Just remember:

> Note that this branch currently switches from doubling the Algorand account as the wallet ID to doubling the XRPL account as wallet ID: see key_from_id. We should change this before merging.

PR #179, opened 2022-04-26
refactor: widen SignTransaction schema + operations

PiDelport:
This makes space for multiple transaction types, rather than just Algorand.

### Related

- Precedes https://github.com/ntls-io/nautilus-wallet/pull/176


longtomjr (approved):
Looks good!

PR #181, opened 2022-04-27
test(web-client,PayPage): increase Storybook load timeout

PiDelport:
2 seconds still seems to hit the timeout sometimes.


IscoRuta98 (approved):
Looks good to go.

jongbonga (approved):

PR #198, opened 2022-05-03
feat(web-client): add opportunistic XRPL token opt-ins

PiDelport:
This performs opportunistic XRPL token opt-in to any peer trustlines on wallet open (similar to the Algorand ASA opt-in).

This needs more proper UI at some point, but this is a minimal foundation to enable basic XRPL token transacting.

### DIFF
--- a/web-client/src/app/state/session-xrpl.service.ts
+++ b/web-client/src/app/state/session-xrpl.service.ts
@@ -12,6 +12,7 @@ import { withLoggedExchange } from 'src/app/utils/console.helpers';
 import { panic } from 'src/app/utils/errors/panic';
 import { TransactionSigned, TransactionToSign } from 'src/schema/actions';
 import * as xrpl from 'xrpl';
+import { Trustline } from 'xrpl/dist/npm/models/methods/accountLines';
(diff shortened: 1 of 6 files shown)


web-client/src/app/state/session-xrpl.service.ts
```diff
@@ -77,6 +88,85 @@ export class SessionXrplService {
...
+
+  /**
+   * Helper: Opt in to the given trust-line.
+   *
+   * This sends a `TrustSet` transaction matching the peer's limit
+   * if the active session's wallet's limit is zero.
+   *
+   * @return the `TrustSet` response, or undefined
+   */
+  async checkTrustlineOptIn(
```
longtomjr, line 145: Name and docs don't match up

longtomjr (approved):
Looks good

web-client/src/app/state/session-xrpl.service.ts
PiDelport: Good catch!

Should be resolved by: [0f27d67..c497b29e](https://github.com/ntls-io/nautilus-wallet/compare/0f27d67..c497b29e)

Issue #212, opened 2022-05-09
Payment Page Blank


Claire-Bear:
After entering in a wallet address I receive a blank page with no more options or feedback to the user.
- validation messages need to be displayed to user
- Page should provide user with fields to make a payment to that wallet of X amount and confirm payment

PR #221, opened 2022-05-10
fix(web-client): prevent payments when the user has no balances

longtomjr:
Related to:
- #212

Closes: #212

### DIFF
--- a/web-client/src/app/views/wallet/wallet.page.ts
+++ b/web-client/src/app/views/wallet/wallet.page.ts
@@ -1,6 +1,6 @@
 import { Component, Input, OnInit } from '@angular/core';
 import { LoadingController, ToastController } from '@ionic/angular';
-import { map, Observable } from 'rxjs';
+import { combineLatest, map, Observable } from 'rxjs';
 import { checkTxResponseSucceeded } from 'src/app/services/xrpl.utils';
 import { SessionAlgorandService } from 'src/app/state/session-algorand.service';
 import { SessionXrplService } from 'src/app/state/session-xrpl.service';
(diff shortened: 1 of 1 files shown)


web-client/src/app/views/wallet/wallet.page.ts
```diff
@@ -39,13 +39,16 @@ export class WalletPage implements OnInit {
   balances: Observable<AssetAmount[]> = this.sessionQuery.allBalances;
 
   /** Enable the "Send Money" action if KYC status is either cleared or not required. */
```
PiDelport, line 41:
```suggestion
  /**
   * Enable the "Send Money" action if both:
   * - KYC status is either cleared or not required
   * - At least one balance is available
   */
```

web-client/src/app/views/wallet/wallet.page.ts
```diff
@@ -39,13 +39,16 @@ export class WalletPage implements OnInit {
...
-  actionSendMoneyEnabled: Observable<boolean> =
-    this.sessionQuery.onfidoCheckIsClear.pipe(
-      map(
-        (onfidoCheckIsClear: boolean) =>
-          onfidoCheckIsClear || !this.requireKycBeforeSendPayment
-      )
-    );
+  actionSendMoneyEnabled: Observable<boolean> = combineLatest(
+    this.sessionQuery.onfidoCheckIsClear,
+    this.sessionQuery.allBalances
```
PiDelport, line 44: It might be good to rather use `this.balances` here to explicitly align with the displayed balances, in case we later start implementing functionality to filter and show only show a subset of all available balances?

```suggestion
    this.balances
```

PiDelport (approved):
Looks good.

web-client/src/app/views/wallet/wallet.page.ts
longtomjr: I thought about using `this.balances` instead. The reason I ended up going with the one from state is because we want a default balance shown if there is no balance in `this.sessionQuery.allBalances` and the way to implement that might be to set a default value for `this.balances`. 

If we later want to have some filtered balances I also think that it might be better add a query for it on `SessionQuery` instead, and update the usage.

Issue #211, opened 2022-05-09
Retain Session on app reopen


Claire-Bear:
If a user navigates away on mobile web and returns to the app (E.g. They nav to whatsapp to copy a wallet address to pay and navigate back into the mobile web app) the user gets an error and is logged out of their session.

Possibly related to the connection closing issue on desktop if you leave your wallet open and unattended for too long.

Brianspha:
I found the cause of the issue it turns out IOS doesn't allow any apps to make any HTTP requests whilst the app is in the background, when I inspected the console the ripple library seems to be the culprit which results in this error being thrown ``WebSocket connection to 'wss://s.altnet.rippletest.net/' failed: The operation couldn’t be completed. (kNWErrorDomainPOSIX error 53 - Software caused connection abort)
``   looking through several blogs I came across this: https://developer.apple.com/forums/thread/106838 its not a solution but  it gives a bit of context

jongbonga:
> IOS doesn't allow any apps to make any HTTP requests whilst the app is in the background

We can, we need to activate the background fetch & processing. I will do that and see if it solved the issue

PR #223, opened 2022-05-10
chore(web-client): enable state persistence for demo deployment

longtomjr:
Closes: #211


PiDelport (approved):
Looks good.

We may want to start creating dedicated environment configurations for the various deployments (`demo.ntls.io`, CBDC sandbox, etc.), instead of just having an undistinguished "production" deployment, but we can do that in follow-up.

Issue #219, opened 2022-05-10
Log out does not work


Claire-Bear:
Log out does not do anything.

PiDelport:
This is probably a duplicate of:

- https://github.com/ntls-io/nautilus-wallet/issues/209

Issue #209, opened 2022-05-06
Reset session state on logout


PiDelport:
Currently, the logout button just links to the landing page view, which implicitly reset the state, but the header component itself should actually take care of clearing the session state.

PR #225, opened 2022-05-10
fix(web-client): clear stores on logout action

longtomjr:
Closes: #219

### DIFF
--- a/web-client/src/app/components/header/header.component.ts
+++ b/web-client/src/app/components/header/header.component.ts
@@ -1,4 +1,6 @@
 import { Component, Input, OnInit } from '@angular/core';
+import { resetStores } from '@datorama/akita';
+import { NavController } from '@ionic/angular';
 import { SessionQuery } from 'src/app/state/session.query';
 
 @Component({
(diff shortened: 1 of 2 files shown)


web-client/src/app/components/header/header.component.ts
```diff
@@ -8,7 +10,15 @@ import { SessionQuery } from 'src/app/state/session.query';
...
 export class HeaderComponent implements OnInit {
   @Input() title = 'NAUTILUS';
-  constructor(public sessionQuery: SessionQuery) {}
+  constructor(
+    public sessionQuery: SessionQuery,
+    private navController: NavController
+  ) {}
+
+  logoutAction() {
+    checkResetStores();
```
PiDelport, line 19: I think we specifically want `resetStores` here, not `checkResetStores`: it's precisely when `persistAkitaState` is enabled that we really want this action to clear the session.

PiDelport (changes requested):

PiDelport (approved):
Looks good!

longtomjr:
Closed: #209

PR #229, opened 2022-05-10
fix(web-client): update balance size

jongbonga:


Claire-Bear:
Please provide a PR description 

What the change was and why it was made but otherwise I approve.

billguo99 (approved):
Ship it! :)

PiDelport:
Note that this no longer vertically right-aligns the amounts: is that something we want?

jongbonga:
> Note that this no longer vertically right-aligns the amounts: is that something we want?

Looking into it now

jongbonga:
> It also truncates large amounts with "…" instead of wrapping. We should maybe consider a tooltip or similar for that case, as a follow-up?

Yes, but we should get at least a 100 billion before the truncate.

Issue #224, opened 2022-05-10
Mobile Text Overlap 


Claire-Bear:
The refresh and logout buttons overlap with the nautilus logo in the middle on mobile.

Perhaps we can put these together in a hamburger menu that slides in from the right. (Open to suggestions).

PiDelport:
This starts happening on extremely narrow (<360 px wide) viewports.

A good way to approach fixing this would be to first define some Storybook stories for the header component at key breakpoint sizes.

jongbonga:
@Claire-Bear we can use icons only buttons

longtomjr:
Closed with: https://github.com/ntls-io/nautilus-wallet/issues/224

PR #234, opened 2022-05-11
Mobile Text Overlap #224

Brianspha:
### Rational 

I decided to not use the hamburger approach to resolving this issue as it was a bit of a challenge getting the ``<ng-content></ng-content>`` to appear in line with the hamburger for screens with width less than ``260px`` so the approach I took was to hide the ``NAUTILUS`` logo/title each time we hit the less than ``260px`` screen width break point see screenshots.

normal
small


.gitignore
```diff
@@ -0,0 +1,4 @@
+
+.vscode/settings.json
+yarn.lock
```
jongbonga, line 3: we shouldn't have yarn since we are using npm in this project

web-client/src/app/components/header/header.component.scss
```diff
@@ -0,0 +1,8 @@
+#titleHeaderLabel {
+  display: block;
+}
+@media screen and (max-width: 260px) {
```
jongbonga, line 4: Since we are using tailwind, we can [extend](https://tailwindcss.com/docs/screens#adding-smaller-breakpoints) the breakpoints to the desired size and apply the class to the element

web-client/src/app/views/wallet/wallet.page.html
```diff
@@ -1,7 +1,9 @@
 <ion-header class="ion-no-border">
   <app-header>
     <ion-buttons slot="end">
-      <ion-button (click)="onRefresh()">Refresh</ion-button>
+      <ion-button (click)="onRefresh()">
```
jongbonga, line 4: When a button doesn't have text, it's always a good idea to add `icon-only` to its [slot](https://ionicframework.com/docs/api/button#slots)

PR #235, opened 2022-05-11
Mobile text overlap #224

Brianspha:
### Rational 

I decided to not use the hamburger approach to resolving this issue as it was a bit of a challenge getting the ``<ng-content></ng-content>`` to appear in line with the hamburger for screens with width less than ``260px`` so the approach I took was to hide the ``NAUTILUS`` logo/title each time we hit the less than ``260px`` screen width break point see screenshots.

normal
small


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "makefile.extensionOutputFolder": "./.vscode"
```
jongbonga, line 2: what does this do?

Brianspha: I removed this

longtomjr (approved):
Looks good to me thanks. This can go in once everything has been resolved.

jongbonga (approved):

jongbonga:
My only concern is the buttons are next to each other and it's easy to unintentionally click on the logout.
Maybe we could have the refresh on  the far left `slot="start"`

Issue #215, opened 2022-05-09
No Validation on Mobile number of sign up


Claire-Bear:
- On the mobile field, I can't input "+" for an international number. When I use the alternative by placing "0027" I run out of number length
-  I can still create an account with a bogus number. I placed (999) 999-99-99

Claire-Bear:
Validation should be shown to the user on invalid number formats

PiDelport:
Relevant code:

https://github.com/ntls-io/nautilus-wallet/blob/1772a8bc65dc439b6ed131a5207be390f2f0e601/web-client/src/app/views/register/register.page.ts#L25-L30

https://github.com/ntls-io/nautilus-wallet/blob/1772a8bc65dc439b6ed131a5207be390f2f0e601/web-client/src/app/state/session.service.ts#L165-L167

PiDelport:
As per the comments, we almost certainly want to replace this with something `libphonenumber` based that parses and validates the number to E.164 at entry time.

IscoRuta98:
@PiDelport, would somehting like this: [How to add Angular phone number format](https://edupala.com/how-to-add-angular-phone-number-format/), work?

PiDelport:
Maybe, though that seems to use the heavier-weight `google-libphonenumber` port, rather than `libphonenumber-js`. I'm also wary of that widget's UI styling not necessarily working with our app; particularly, the fact that it seems to be based on `ngx-bootstrap`.

Googling a bit more, maybe this is worth a try? https://github.com/azzamasghar1/ion-intl-tel-input

IscoRuta98:
Thanks Pi!
Checking it now. Quick question: are the forms we use template-driven or reactive?

PR #242, opened 2022-05-13
feat(web-client):  add country code & validation to register page

jongbonga:
Allow to format phone number to have prefix


PiDelport:
ion-intl-tel-input is probably a better approach here, if we can get the styling fixed?

- https://github.com/ntls-io/nautilus-wallet/pull/231

Regardless of UI choice, the crucial thing is that we need the numbers normalised to E.164, so we'll almost certainly want to use libphonenumber; see <https://github.com/ntls-io/nautilus-wallet/issues/215#issuecomment-1122294364>.

jongbonga:
> ion-intl-tel-input is probably a better approach here, if we can get the styling fixed? See this PR:
> 
> * [feat(web-client): use ion-intl-tel-input for mobile number input #231](https://github.com/ntls-io/nautilus-wallet/pull/231)
> 
> Regardless of UI choice, the crucial thing is that we need the numbers normalised to E.164, so we'll almost certainly want to use libphonenumber; see [#215 (comment)](https://github.com/ntls-io/nautilus-wallet/issues/215#issuecomment-1122294364).

updated to use the library

web-client/src/app/app.module.ts
```diff
@@ -27,6 +28,7 @@ import { TranslocoRootModule } from './transloco/transloco-root.module';
     AkitaNgRouterStoreModule,
     NgxPrinterModule.forRoot({ printOpenWindow: false }),
     ErrorHandlerModule,
+    IonIntlTelInputModule,
```
longtomjr, line 31: Do we need to import this in the app.module and the register.module?

longtomjr (approved):
Looks good.

Maybe just change this section in the session service as well, since I don't think it is needed anymore.
https://github.com/ntls-io/nautilus-wallet/blob/68c5358cbde2cdacb05abfb5786237fcf78c8ae7/web-client/src/app/state/session.service.ts#L164-L167

web-client/src/app/app.module.ts
jongbonga: No, thanks!

PR #249, opened 2022-05-17
fix(web-client): Fix register UI & Pin validation

jongbonga:
- Remove validation underlines on inputs.

- Validation was ignored after the pin was entered, highlighted then deleted.

- It also addresses the case where confirm pin does not validate after the match happened but the pin is deleted after.
- Add `minlength` for FirstName & LastName


Claire-Bear:
Task linked: [CU-2au94gv Validation on Sign Up Pin](https://app.clickup.com/t/2au94gv)

longtomjr (approved):

PR #265, opened 2022-06-01
feat(web-client): show transaction fees

jongbonga:
Display transaction fees & total to be sent after the fee on the pay page


longtomjr (changes requested):
For consistency I think we should continue using the module for each component.

longtomjr (approved):

PR #326, opened 2022-07-25
feat(web-client): OTP UI prompt

jongbonga:


billguo99:
Looks good, will revisit once the backend functionality is integrated.

PR #333, opened 2022-08-01
build(web-client): update dependencies

jongbonga:


billguo99 (approved):

billguo99:
If we don't keep our dependencies up to date, what could happen worst case scenario? :)

jongbonga:
> If we don't keep our dependencies up to date, what could happen worst case scenario? :)

Security issues, dependencies mismatch, missing new features

PR #335, opened 2022-08-05
feat(web-server): Implement OTP functionality.

IscoRuta98:
## Implementing 2FA 

- Description of the feature: https://app.clickup.com/t/2cyrmaz
- The aim of the PR is implementing the server-side feature of 2FA by prompting users to give OTP which is sent form the server via Twilio' SMS feature.

### Tasks

- [x] Implement a GenerateOTP function
- [x] Store OTP in Wallet storable.
- [x]  Add OTP verification before transaction is signed.
- [x] Connect web-server OTP to the client-side.
- [x] Use existing Twilio messaging service (from the client-side) to send OTP.
- [ ] Implement server-side tests for generate_otp() and sign_transaction_with_otp() function.

### UI Flow

1. User initiates a payment.
2. If payment is above a certain limit, generateOtp() function is triggered, were the user is sent an OTP (via Twilio) and an OTP popup is displayed.
3. If user gets OTP correct, transaction is sent to the enclave to be signed (i.e. signTransactionWithOtp is triggered) and posted on the blockchain ledger.


web-server/sgx-wallet-impl/src/schema/actions.rs
```diff
@@ -188,6 +194,8 @@ impl From<UnlockWalletError> for SaveOnfidoCheckResult {
         match err {
             InvalidWalletId => Self::InvalidAuth,
             InvalidAuthPin => Self::InvalidAuth,
+            //FIXME: Do additional error handling to prevent panic.
+            InvalidOtp => unreachable!(), // This variant should never happen as otp is not a feature required for KYC.
```
billguo99, line 198: I don't think this type of error is needed here as OTP errors will only occur in Payment transactions.

web-server/sgx-wallet-impl/src/schema/actions.rs
```diff
@@ -221,6 +242,8 @@ impl From<UnlockWalletError> for LoadOnfidoCheckResult {
...
```
billguo99, line 246: Think we can remove this as I think OTP errors will only occur in Payment transactions.

web-server/sgx-wallet-impl/src/schema/actions.rs
```diff
@@ -52,6 +52,7 @@ impl From<UnlockWalletError> for OpenWalletResult {
         match err {
             InvalidWalletId => Self::InvalidAuth,
             InvalidAuthPin => Self::InvalidAuth,
+            InvalidOtp => Self::InvalidAuth,
```
billguo99, line 55: Don't think this is needed here. We are not using OTP when opening the wallet.

web-server/sgx-wallet-impl/src/schema/actions.rs
```diff
@@ -66,14 +67,18 @@ pub struct SignTransaction {
 
     #[zeroize(skip)]
     pub transaction_to_sign: TransactionToSign,
+
+    #[zeroize(skip)]
+    pub otp: Option<String>
```
billguo99, line 72: Like in [web-client/src/schema/actions.ts](https://github.com/ntls-io/nautilus-wallet/pull/335/files#diff-545973229f602022dfd359b6af35c03698020e5fca5d49aaa95c247b793484fb): I think we need a separate struct for `SignTransactionWithOtp`.

web-server/sgx-wallet-impl/src/wallet_operations/generate_otp.rs
```diff
@@ -0,0 +1,25 @@
...
+    let number = rand_number.to_string();
+    let otp = "0".repeat(OTP_DIGIT_LENGTH-number.len()) + &number;
+
+    let store_otp = | mut stored:WalletStorable | {
+
+        stored.otp = Some(otp);
+        stored
+    };
+    GenerateOtpResult::Generated(mutate_wallet(&request.wallet_id, store_otp).unwrap().unwrap().otp.clone().unwrap())
+}
```
billguo99, line 25: minor fix: need a new line at end of file :)

billguo99 (changes requested):
Still need some tests for the `generate_otp` function in sgx-wallet-test.

web-client/src/app/state/session.service.ts
```diff
@@ -140,6 +170,47 @@ export class SessionService {
...
 
+  async signTransactionWithOtp(
+    transaction_to_sign: TransactionToSign
+  ): Promise<TransactionSigned> {
+    const { wallet, pin } = this.sessionQuery.assumeActiveSession();
+
+    const signRequest: SignTransactionWithOtp = {
+      auth_pin: pin,
+      wallet_id: wallet.wallet_id,
+      transaction_to_sign,
```
billguo99, line 181: We need to add OTP to the `signRequest`.

web-client/src/schema/actions.ts
```diff
@@ -23,13 +23,27 @@ export type OpenWalletResult =
...
   wallet_id: WalletId;
   auth_pin: WalletPin;
 
   transaction_to_sign: TransactionToSign;
 };
 
+export type SignTransactionWithOtp = {
+  wallet_id: WalletId;
+  auth_pin: WalletPin;
+  /** otp: Otp; */
```
billguo99, line 42: We should uncomment this line as we need the otp in this type.

web-server/sgx-wallet-impl/src/schema/actions.rs
billguo99: We can make this just a `String` and remove the `Option` so we ensure that there has to be an OTP.

web-server/sgx-wallet-impl/src/schema/actions.rs
IscoRuta98: When I don't include Invalidotp in the `match err`, I get the following
error: **_error[E0004]:_ non-exhaustive patterns: InvalidOtp not covered_**

In my latest commit, I've added a todo()! macro for now.

billguo99: You get the error because the enum `UnlockWalletError` has a arm for `InvalidOtp` - https://github.com/ntls-io/nautilus-wallet/blob/9529dd471a97995966de42f69c083733ed9664c2/web-server/sgx-wallet-impl/src/wallet_operations/store.rs#L97.

Think we can just remove that.

billguo99: Actually ignore that, I see that you bundled the OTP functionality with unlocking the wallet - https://github.com/ntls-io/nautilus-wallet/blob/9529dd471a97995966de42f69c083733ed9664c2/web-server/sgx-wallet-impl/src/wallet_operations/store.rs#L64. Therefore you have to have the `InvalidOtp` enum there.

I don't think we should combine unlocking the wallet with the OTP, as the OTP is just used for deciding whether to sign the transaction or not.

billguo99:
Have you tried implementing some tests in `sgx-wallet-test`for generate_otp and sign_transaction_with_otp?

web-server/sgx-wallet-impl/src/schema/actions.rs
IscoRuta98: Should I rather create a `UnlockWalletErrorWithOtp` enum perhaps?
As the `UnlockWalletError` enum is needed when OTP is checked by the user's response.

billguo99: You can or you can just leave as it is for now.

I thought that comparing the OTP wouldn't require the method to  "unlock wallet".

IscoRuta98: I think we should leave it for now.

To compare OTP, would we not need to first unlock the wallet first, then compare OTP that's stored in the wallet against the one provided.

I tried implementing tests but I can't finish it in time. I plan to write `generate_otp` and `sign_transaction_with_otp` tests at a later time.

billguo99: To access the OTP, the enclave doesn't need the pin to access it - https://github.com/ntls-io/nautilus-wallet/blob/9529dd471a97995966de42f69c083733ed9664c2/web-server/sgx-wallet-impl/src/wallet_operations/store.rs#L66.

But let's see how this way works first.

The tests are important to check/ensure that they work as intended. If we skip the tests, we only check if these backend methods work when we integrate it with the frontend.

PR #324, opened 2022-07-25
feat(server): initial implementation of PIN reset functionality

jdvlio:
### Overview

A wallet user that has forgotten their PIN should be allowed to reset it after answering one or more security questions.  These security questions and their corresponding answers are set upon account creation.  The enclave code authenticates all such reset requests using the valid responses, which were sealed and stored upon account initialization.

### Task Breakdown

The following functionality should be implemented:
- [x] A new type to facilitate the creation, ~modification~ and validation of security responses
- [x] Valid security responses should be sealed as part of the `WalletStorable` struct
- [x] Policy under which PIN resets should be authorized
- [ ] ~User security questions should be retrievable as plaintext via a sensible mechanism~
- [ ] Update backend integration tests
- [ ] Connect with frontend

### Client Functionality Required

The frontend should define the security questions users may choose from.  The backend will allow arbitrary security questions which will be retrievable via HTTP during the reset process.

The functionality in this PR requires two separate exchanges, via MessagePack, between the client and server:

#### Initial request

Name: **StartPinReset**
Request:

| Field | Type | Encoding | Description |
| - | - | - | - |
| wallet_id | string | UTF-8 | User's wallet id. | 
|  wallet_auth_map | Map<string,string> | UTF-8 (keys and values) | Security questions and attempts. |
|  client_pk| uint8[32] | - | Client public key for a newly created session. |

Response variant A:

| Field | Type | Encoding | Description |
| - | - | - | - |
| ServerPk | uint8[32] | ~base64~ | Enclave public key for this session. |

Response variant B:

| Field | Type | Encoding | Description |
| - | - | - | - |
| InvalidAuth | null | - | Invalid authentication details received. |

Response variant C:

| Field | Type | Encoding | Description |
| - | - | - | - |
| Failed | string | UTF-8 | Error message: unexpected error encountered |

#### Follow-up request

Name: **PinReset**
Request:

| Field | Type | Encoding | Description |
| - | - | - | - |
| wallet_id | string | UTF-8 | User's wallet id. | 
| new_pin | string | UTF-8 | User's new PIN. |
| new_pin_mac | uint8[32] | ~base64~| MAC of the new PIN. | 
|  client_pk| uint8[32] | - | Client public key for a newly created session. |

Response variant A:

| Field | Type | Encoding | Description |
| - | - | - | - |
| Reset | null | - | User PIN succesfully reset. |

Response variant A:

| Field | Type | Encoding | Description |
| - | - | - | - |
[...]

PR #344, opened 2022-08-19
[WIP] feat(web-server): initial wallet enclave sessions

jdvlio:
### Overview

Certain multi-step wallet operations require that two or more messages related messages be sent for processing by the enclave.   Since the enclave has no built-in means of verifying that such messages are indeed from the same source, a cryptographic mechanism is necessary to essentially establish "sessions" across such messages.

### Aim
 
Provide a simple, albeit rudimentary, solution to the aforementioned problem through the use of randomly generated session keys and using keyed hashes of prior messages as a session identifier.

### Context

The need for this functionality arose in the context of PR #324, where authentication details need to be verified prior to a follow-up message that contains the new to-be PIN for the user wallet.  The enclave needs some mechanism by which it can verify that the latter message originates from the same source as the former containing the user responses to pre-defined security questions.

### DIFF
--- a/web-server/sgx-wallet-impl/Cargo.lock
+++ b/web-server/sgx-wallet-impl/Cargo.lock
@@ -132,6 +132,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.10.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


billguo99 (approved):
https://www.youtube.com/watch?v=MKn3cxFNN1I - useful YT video to understand how HMAC confirms message comes from a certain party that has the same secret as you :)

PR #351, opened 2022-09-05
feat: initial wallet enclave sessions

jdvlio:
### Overview

Certain multi-step wallet operations require that two or more messages related messages be sent for processing by the enclave.   Since the enclave has no built-in means of verifying that such messages are indeed from the same source, a cryptographic mechanism is necessary to essentially establish "sessions" across such messages.

### Aim
 
Provide a simple, albeit rudimentary, solution to the aforementioned problem through the use of a combination of a Diffie-Hellman key exchange and session keys deterministically generated by the enclave.

### Context

The need for this functionality arose in the context of PR #324, where authentication details need to be verified prior to a follow-up message that contains the new to-be PIN for the user wallet.  The enclave needs some mechanism by which it can verify that the latter message originates from the same source as the former containing the user responses to pre-defined security questions.

### DIFF
--- a/web-client/src/schema/crypto.ts
+++ b/web-client/src/schema/crypto.ts
@@ -19,7 +19,7 @@ type EncryptedMessage = {
 };
 
 export class TweetNaClCrypto {
-  constructor(public keyPair: BoxKeyPair) { }
+  constructor(public keyPair: BoxKeyPair) {}
 
   get public_key(): PublicKey {
     return this.keyPair.publicKey;
--- a/web-client/src/schema/crypto.ts
+++ b/web-client/src/schema/crypto.ts
@@ -19,7 +19,7 @@ type EncryptedMessage = {
 };
 
 export class TweetNaClCrypto {
-  constructor(public keyPair: BoxKeyPair) {}
+  constructor(public keyPair: BoxKeyPair) { }
 
   get public_key(): PublicKey {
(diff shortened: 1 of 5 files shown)


web-client/src/schema/session.ts
```diff
@@ -0,0 +1,57 @@
...
+  constructor(protected dh: DiffieHellman, readonly theirPk: PublicKey) {
+    this.secret = dh.diffie_hellman(theirPk);
+    this.ourPk = dh.x25519_public_key();
+  }
+
+  /**
+   * Compute and return a message authentication code for an input message.
+   */
+  message_authentication_code = (msg: Bytes): Bytes32 => {
+    let hmac = new HMAC(SHA256, this.secret);
```
jongbonga, line 53:
```suggestion
    const hmac = new HMAC(SHA256, this.secret);
```

web-client/src/schema/crypto.ts
```diff
@@ -19,7 +19,7 @@ type EncryptedMessage = {
 };
 
 export class TweetNaClCrypto {
-  constructor(public keyPair: BoxKeyPair) {}
+  constructor(public keyPair: BoxKeyPair) { }
```
billguo99, line 22: npm run format

billguo99 (changes requested):

billguo99 (approved):

PR #352, opened 2022-09-12
feat(web-client): add pull payment page

jongbonga:
Pull payments steps 

https://docs.google.com/document/d/1ni7n0RLirbTDDHtqZZ7KZLVDQVPUoZuznc__zQK3JLI/edit?pli=1

https://app.clickup.com/t/2ejcc5h

Manually tested the following:
* Pull payments for XRP and for tokens and commission pull payments for both XRP and for tokens
* Also tested the hideXRP flag feature 

Known issues:
* Scan a QR code button just shows a camera


web-client/src/app/views/pay/pay.page.html
```diff
@@ -9,7 +9,7 @@
 <ion-content>
   <app-pure-pay-page
     [senderName]="senderName | async"
-    [receiverAddress]="receiverAddress | async"
+    [receiverAddress]="receiverAddress"
```
billguo99, line 12: what does removing `| async` do over here?

jdvlio: See https://angular.io/api/common/AsyncPipe#description

web-client/src/app/state/session-xrpl.service.ts
```diff
@@ -171,10 +180,21 @@ export class SessionXrplService {
     }
     const mainTxnUnsigned: xrpl.Payment = await this.prepareUnsignedTransaction(
       receiverId,
-      mainAmount
+      mainAmount,
+      senderId
+    );
+    let wallet_hex_key;
+    if (senderId) {
+      wallet_hex_key = await this.sessionService.getXrplWalletPublicKey(
```
sonasi, line 188: If I enter the wrong PIN I am automatically logged out with an "Unexpected Error Occurred" message. Should this be the correct behaviour?

image

web-client/src/app/views/pull/pull.page.html
```diff
@@ -0,0 +1,98 @@
...
+          [autofocus]="true"
+          [balance]="balance"
+          [maxAmount]="maxAmount"
+        ></app-pay-amount-form>
+      </ion-col>
+    </ion-row>
+  </ion-grid>
+</ion-content>
+
+<ion-modal [isOpen]="isPinEntryOpen" (didDismiss)="isPinEntryOpen = false">
```
sonasi, line 80: This is more of a design question, so maybe @Claire-Bear could chime in, but could we clarify when asking for the PIN that we are asking for the sender's PIN?
image

web-client/src/app/state/session-xrpl.service.ts
billguo99: Not sure, what do you suggest the correct behaviour to be?

sonasi: It should say 'Incorrect PIN' rather than 'Unexpected Error Occurred' at least. As far as automatically logging the receiver out when the sender puts in the wrong pin, I suppose we need to limit repeated attempts, but one try seems harsh. If it's easy to do we could allow a two or three attempts before logout.

billguo99: I've updated the error handling to display "Incorrect PIN" and not log them out so that they stay on the pull page.

The only issue is that if the autofund ever has the incorrect pin, when a user creates a wallet and the autofund experiences an "invalid auth", then the user will be redirected to the /pull page where they won't be able to do anything.

I think we can implement the two/three attempts before logout as a separate PR if we need it.

web-client/src/app/views/pull/pull.page.html
billguo99: Let's get this feature in and get some input from Bronwynne during staging testing.

web-client/src/app/state/session-xrpl.service.ts
sonasi: image
The unset pin looks good. If the wallet address is wrong it also gives an "unknown error occurred". We could just say "wallet address not found". However, we could open a separate issue for that for expediency.

billguo99: Ah, ok we should add validation check for valid addresses rather. Let's do that as a separate PR.

sonasi (approved):

web-client/src/app/utils/errors/global-error-handler.ts
```diff
@@ -19,6 +19,25 @@ export class GlobalErrorHandler implements ErrorHandler {
       return;
     }
 
+    if (
+      error.message.includes('SessionService.signTransaction: invalid auth')
+    ) {
+      console.warn('GlobalErrorHandler:', error);
+      this.zone.run(
```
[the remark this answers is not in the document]
billguo99: I was trying to copy how we threw an error notification for "Unexpected Error Occurred".

I think we're using zone so that we can use async functions.

If not, can you recommend how I should refactor this?
It's also difficult to move this notification to `session.service.ts` because then the functions return type expects a promise etc.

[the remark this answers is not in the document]
jongbonga: I don't think ngZone is necessary but it doesn't do harm either.
we can have a quick session after the planning to see what can be done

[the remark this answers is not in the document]
billguo99: I caught it somewhere else and threw a Local Error and ignored the local error in the global error handler :)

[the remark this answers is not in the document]
jongbonga: @billguo99 can you just add a TODO to revisit and refactor later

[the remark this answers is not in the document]
billguo99: Done :+1:

PR #370, opened 2022-10-12
feat(web-client): add environment files for both staging and prod for both pilots

billguo99:
Adding these environment files for our new pilots.


IscoRuta98 (approved):
All looks good!

PR #376, opened 2022-10-18
ci(web-client): remove chromatic

billguo99:
It's time to stop the emails for a tool we don't use :)


jongbonga:
Did you already remove it from github actions?

jongbonga (approved):

billguo99:
> @billguo99 Did you already remove it from github actions?

The github actions reads it from https://github.com/ntls-io/nautilus-wallet/blob/main/.github/workflows/web-client.yaml file so after this PR merges to main, it should remove it.

billguo99:
I've edited the branch protection rules to exclude `Storybook Publish` and `Chromatic` checks to not be required.

PR #395, opened 2022-10-27
fix(web-server): update command 'bindgen' to 'bindgen-cli'

billguo99:
`bindgen` recently split out their cli-tool and the library into two parts.
This is causing our web-server tests to fail.


jdvlio (approved):
Perfect! :+1:

PR #410, opened 2022-11-02
feat(web-client): add search service

billguo99:
Adding search service to connect to the 2 backend APIs that connects to our CosmosDB. Also add wallet data into db when new wallet is created. Tested on ntls-staging.

### DIFF
--- a/web-client/src/app/state/session.service.ts
+++ b/web-client/src/app/state/session.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import { EnclaveService } from 'src/app/services/enclave/index';
 import { MessagingService } from 'src/app/services/messaging.service';
+import { SearchService } from 'src/app/services/search.service';
 import { SessionQuery } from 'src/app/state/session.query';
 import { withLoggedExchange } from 'src/app/utils/console.helpers';
 import { panic } from 'src/app/utils/errors/panic';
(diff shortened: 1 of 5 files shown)


Claire-Bear:
Task linked: [CU-315ghwb Add search frontend code](https://app.clickup.com/t/315ghwb)

IscoRuta98 (approved):
LGTM!

web-client/src/app/state/session.service.ts
```diff
@@ -58,6 +60,15 @@ export class SessionService {
...
         const wallet = result.Created;
         this.sessionStore.update({ wallet, pin });
+        try {
+          await this.searchService.insertWalletAddress({
+            wallet_id: wallet.wallet_id,
+            phone_number: wallet.phone_number,
+            owner_name: wallet.owner_name,
+          });
+        } catch (err) {
+          console.log(err);
```
jdvlio, line 70: I believe this should be `console.error`

jdvlio (changes requested):

web-client/src/app/state/session.service.ts
billguo99: No, there are other places in our code that has console.log(err) - https://github.com/ntls-io/nautilus-wallet/blob/23a328a827c853028880e7222da5a5487246cbbf/web-client/src/app/views/register/register.page.ts#L109

sonasi (approved):

web-client/src/app/state/session.service.ts
jdvlio: I see.  I won't pretend to understand it but I assume it relates to how actionable the error message is.

jdvlio (approved):
Seems good

PR #426, opened 2022-11-10
Search wallet frontend

IscoRuta98:
Aim: Connect Search Wallet service to the frontend.

[Clickup Task](https://app.clickup.com/t/2ejcbat)

### DIFF
--- a/web-client/src/app/views/search-wallet/search-wallet.page.html
+++ b/web-client/src/app/views/search-wallet/search-wallet.page.html
@@ -6,6 +6,7 @@
   </ion-toolbar>
   <ion-toolbar>
     <div class="ion-text-center p-2">
+      <ion-icon slot="start" name="wallet" size="large"></ion-icon>
       <h1 class="font-semibold font-nasalization">Search wallet address</h1>
     </div>
   </ion-toolbar>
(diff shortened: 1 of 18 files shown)


web-client/src/app/views/search-wallet/search-wallet.page.html
```diff
@@ -0,0 +1,68 @@
...
+  <ion-grid>
+    <ion-row>
+      <ion-col>
+        <ion-list lines="none" class="form-item">
+          <ion-item>
+            <ion-input placeholder="Name" [(ngModel)]="name"></ion-input>
+          </ion-item>
+          <ion-item>
+            <ion-input
+              placeholder="Phone Number"
```
sonasi, line 24: The name and phone number must be exactly right, including the plus sign before the country code. I could only get it to work by copying and pasting from the database. We could implement a validation where it doesn't let you search unless the number is the right format. Something easier might be just to add to the form cell placeholder with the format. 

I.e. instead of "Phone Number", maybe "Phone Number (+05555555555)".

sonasi (approved):

web-client/src/app/views/search-wallet/search-wallet.page.html
IscoRuta98: Thanks for the suggestion. Would the below changes suffice?
image

sonasi: Yes that is better. We could also reuse the same form validation that is used for phone numbers during the create wallet process.

However, I think we need to add validation to several other places, so maybe we can do that as a separate task.

PR #429, opened 2022-11-15
Firebase Setup values

jongbonga:
Retrieve and use values from the setup page


web-client/src/app/state/setup/setup.store.ts
```diff
@@ -0,0 +1,26 @@
+import { Injectable } from '@angular/core';
+import { Store, StoreConfig } from '@datorama/akita';
+
+export interface SetupState {
+  code: string;
```
billguo99, line 5: need to change this to `tokenSymbol`

web-client/src/app/state/setup/setup.store.ts
```diff
@@ -0,0 +1,26 @@
...
+  tokenIssuer: string;
+  xrpIssuer: string;
+  ledger: string;
+  name: string;
+}
+
+export const createInitialState = (): SetupState => ({
+  code: '',
```
billguo99, line 13: need to change this to `tokenSymbol`

web-client/src/app/state/setup/setup.query.ts
```diff
@@ -0,0 +1,17 @@
...
+import { SetupState, SetupStore } from './setup.store';
+
+@Injectable({ providedIn: 'root' })
+export class SetupQuery extends Query<SetupState> {
+  tokenIssuer = this.getValue().tokenIssuer.trim() || environment.tokenIssuer;
+  xrpIssuer = this.getValue().xrpIssuer.trim() || environment.xrpIssuer;
+  ledger = this.getValue().ledger.trim()
+    ? { ...environment.xrplClient, server: this.getValue().ledger }
+    : environment.xrplClient;
+
```
billguo99, line 13: Will we add other fields as needed like `name`, `tokenSymbol`? Right now, these 3 values are the only ones that needs to be "configurable"?

billguo99 (changes requested):
Nice!

web-client/src/app/state/setup/setup.query.ts
jongbonga: Correct.
The name is already in the environment so shouldn't be configurable.
The tokenSymbol will probably be used yes

billguo99 (approved):

billguo99:
LGTM

PR #432, opened 2022-11-21
feat: initial bookmarking implementation

jdvlio:
### Overview

Initial bookmarking implementation.  An HTTP `POST` endpoint is exposed at the `/data` location and exactly three possible method types (in the form of JSON) are accepted:

| Method | Action |
|-|-|
| `GetBookmarks` | Retrieve a list of of bookmarks for a specified user |
| `CreateBookmark` | Create a new bookmark for an address specified by the user |
| `DeleteBookmark` | Delete the requested bookmark from a user's collection | 

More granular details of this API, such as the underlying fields and their types, are available vie the Swagger/OpenAPI documentation included as part of this PR, as described below.

### Swagger/OpenAPI documentation

The desired documentation is accessible while an instance of the server is running.  Assuming `poetry` is installed and `poetry install` has already been executed in the root of the `python` project, start the server using either

```shell
poetry run uvicorn web_asgi.main:app
```
or

```shell
poetry shell 
uvicorn web_asgi.main:app
```
Once running, the documentation may be accessed via the `localhost` at [this link](http://localhost:8000/docs).

#### Synopsis

The HTTP endpoint is created using the `fastapi` web framework and relevant data is stored in `mongodb` by means of the `motor` library.  Data received from clients talking to this endpoint is parsed and validated against the predefined schemas by means of the `pydantic` library.  Configuration options for the server are set using the environment and/or a `.env` file in the project root.

#### Python

The following tools and libraries from the `python` ecosystem are used:

| Tool/library | Description |
|-|-|
| `pydantic` | A data validation library that leverages `python` type hints.  It guarantees that subclassed types of its `BaseModel` class conform to a predefined schema upon successful parsing of a JSON object.  Supports a myriad of related and useful features such as JSON Schema, validation of configuration options, dynamic models, secret management etc. |
| `fastapi` | An asynchronous web framework that leverages `pydantic` validation and `python` type hints.  Produces OpenAPI documentation for endpoints created using it. |
| `python-dotenv` | A `python` implementation of the ubiquitous `dotenv` library found in virtually all languages.  It reads environment variables from an environment file, usually one matching the glob `*.env`. |
| `uvicorn` | An application server that supports the [ASGI protocol][asgi-protocol] used by modern `python` web frameworks such as `fastapi`. |
| `tox` | A testing frontend that runs tests in freshly installed environments. |
| `ruff` | A linting tool for Python, written in Rust, that encompasses various features from the well-respected `flake8` linter and its family of plugins. |
| `black` | A code formatting tool that homogenises and prettifies your spaghetti. |
| `isort` | A formatting tool that organizes your imports. |

[asgi-protocol]: https://asgi.readthedocs.io/en/latest/


backend-services/pyproject.toml
```diff
@@ -0,0 +1,84 @@
...
+         {include = "tests"},
+         ]
+
+[tool.poetry.dependencies]
+python = "^3.10"
+motor = "^3.1.1"
+fastapi = "^0.87.0"
+pydantic = "^1.10.2"
+python-dotenv = "^0.21.0"
+uvicorn = {extras = ["standard"], version = "^0.20.0"}
```
sonasi, line 20: Could you write a high level description of these libraries and what they do, also for ruff, black, etc. Two or three paragraphs summarizing how these fit together since it's the first use in our codebase.

backend-services/src/common/types.py
```diff
@@ -0,0 +1,6 @@
+from typing import NewType
+
+WalletAddress = NewType("WalletAddress", str)
```
sonasi, line 3: Is there a benefit to making a type for this over it being a string? Can we enforce, say, length? Or is this just as an example?

backend-services/src/data_service/operations/bookmark.py
```diff
@@ -0,0 +1,68 @@
...
+    CreateBookmarkResult,
+    DeleteBookmark,
+    DeleteBookmarkResult,
+    GetBookmarks,
+    GetBookmarksResult,
+)
+from data_service.schema.entities import Bookmark, BookmarkDocument
+from data_service.settings import MongoSettings
+
+mongo_settings = MongoSettings()
```
sonasi, line 16: I think these should be all caps, e.g. MONGO_SETTINGS, does the linter not complain?

backend-services/src/data_service/operations/bookmark.py
```diff
@@ -0,0 +1,68 @@
...
+    collection = db[settings.wallet_db_bookmark_collection]
+    return collection
+
+
+async def create_bookmark(params: CreateBookmark) -> CreateBookmarkResult:
+    """
+    Create a new bookmark.
+    """
+    collection = bookmark_collection(client, mongo_settings)
+    new_doc = BookmarkDocument.parse_obj(params)
```
sonasi, line 33: I'd say name this `bookmark_doc` as `new_doc` is not very descriptive.

sonasi (approved):

backend-services/pyproject.toml
```diff
@@ -0,0 +1,84 @@
...
+gunicorn = "^20.1.0"
+
+[tool.poetry.group.test.dependencies]
+pytest = "^7.2.0"
+pytest-mock = "^3.10.0"
+mock = "^4.0.3"
+pytest-asyncio = "^0.20.2"
+
+[tool.poetry.group.dev.dependencies]
+mypy = "^0.991"
```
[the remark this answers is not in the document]
jdvlio: Yes.  Its main purpose is to serve as a static type checker locally and in our CI.  The reason I have not incorporated it into our CI yet is that the `motor` library hasn't been properly type annotated by its authors yet.  What this means for us is that I'll first have to create the necessary [stubs][typing-stub].

Once these stubs have been added we'll properly be able to check our code for any type errors before even merging.

[typing-stub]: https://mypy.readthedocs.io/en/stable/stubs.html

backend-services/pyproject.toml
jdvlio: Definitely, not a problem!

backend-services/src/common/types.py
jdvlio: Definitely, you'll especially notice this in our Rust code.   I'll sketch one example here.  Imagine a function which takes `public_key` and `secret_key` as arguments, in that order.  If these two values have different types and the caller mistakenly messes up the order of the arguments then `mypy` (Python) or the compiler (Rust) will create big red messages prompting the caller to rectify their mistake.  If both have the same type then this will go unnoticed and possibly even go into production.

Long story short, languages that support static type checking make it possible to assign types in a manner that corresponds to the semantics of the values it is intended to represent.

backend-services/src/data_service/operations/bookmark.py
jdvlio: The value of `mongo_settings` depends on information known only at runtime and is therefore not constant.  All caps are reserved to values that, essentially, are known at the time of writing the code (compile time if you prefer).

backend-services/README.md
```diff
@@ -0,0 +1,110 @@
...
+In order to proceed, one needs to have the `poetry` dependency management and
+packaging tool installed. Unless a recent version is available via your OS
+package manager, as it would be on Arch Linux and friends :), the recommended
+means of installing `poetry` is via the well-established `pipx` tool as
+described in [their documentation][pipx-install].
+
+Once `pipx` is installed, installing `poetry` is as simple as
+
+```shell
+pipx install poetry
```
billguo99, line 23: JP helped me debug that my poetry was still using Ubuntu's Python 3.8 and we fixed my env by specifying the python version when installing poetry.
```pipx install poetry --python python3.10```

PR #464, opened 2022-12-05
refactor(client): expose client public key

jdvlio:
The client public key was previously unobtainable by the caller.


IscoRuta98 (commented):
Why did you remove the `protected` term on the constructors for `crypto.ts` and `session.ts`?

sonasi (approved):

PR #468, opened 2022-12-06
feat(web-client): Transaction history 

jongbonga:
Show transaction history for the session's wallet address


sonasi (approved):
lgtm for transfers

PR #481, opened 2022-12-12
feat(web-client): add auto logout

billguo99:
https://app.clickup.com/t/31pkyq8

We want to protect the user by forcing logout after specific conditions are met:
1. Time out and log user out after 1min inactive
2. If transacting through a connector, force user to logout after each transaction. Push & pull payments through a connector.

Flaws:
If transaction or operation takes longer than 1 min, auto logout may interfere and cause the transaction to fail. Perhaps this is a good thing as we do not want any operation to take longer than a minute.

image

### DIFF
--- a/web-client/src/app/services/auto-logout.service.ts
+++ b/web-client/src/app/services/auto-logout.service.ts
@@ -12,7 +12,6 @@ const STORE_KEY = 'lastAction';
   providedIn: 'root',
 })
 export class AutoLogoutService {
-  isLogin = false;
   constructor(
     private navCtrl: NavController,
     private sessionStore: SessionStore,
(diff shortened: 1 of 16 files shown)


web-client/src/app/services/auto-logout.service.ts
```diff
@@ -0,0 +1,85 @@
...
+
+      if (isTimeout) {
+        this.cleanUp(true);
+      }
+    }
+  }
+
+  public cleanUp(notification: boolean) {
+    localStorage.removeItem(STORE_KEY);
+    removeEventListener('click', () => this.reset());
```
jongbonga, line 70: why not `removeAllListeners`?

billguo99: Didn't know there was such a concise method :) thanks!

sonasi (approved):

web-client/src/app/services/auto-logout.service.ts
billguo99: I got some errors when calling `removeAllListeners` so I've reverted it back to `removeEventListener`.

PR #482, opened 2022-12-12
fix(web-client): add issuers for servers

sonasi:
Update production pilot issuers.


web-client/src/environments/environment.palau-prod.ts
```diff
@@ -28,19 +28,19 @@ export const environment: Environment = {
   // USDC from https://testnet.folks.finance/faucet
   defaultAlgorandAssetId: 67395862,
   xrplClient: {
-    server: 'wss://s1.cbdc-sandbox.rippletest.net:51233',
+    server: 'wss://s1.ripple.com',
```
jdvlio, line 31: Is it not necessary to specify the port here?  I recall this creating issues for us in the past.
```suggestion
    server: 'wss://s1.ripple.com:51233',
```

jdvlio (changes requested):

web-client/src/environments/environment.palau-prod.ts
sonasi: Evidently this server works without the port for web sockets, however it also works with the port so I'll keep it.

web-client/src/environments/environment.palau-prod.ts
```diff
@@ -28,19 +28,19 @@ export const environment: Environment = {
...
   commissionPercentage: 0.0,
-  tokenIssuer: 'rngdKd8BAM3etQcb12DvGxd5Ps9MocAvPa',
+  tokenIssuer: 'rwekfW4MiS5yZjXASRBDzzPPWYKuHvKP7E',
   tokenSymbol: 'PSC',
-  xrpIssuer: 'rJsc4o9Kh5xx6ba2pR62FbFpYJoh3xetMk',
+  xrpIssuer: 'r3xUVoiRsww3SpPhkYaPfhNZ52K8KP5rEc',
   hideXrpBalance: true,
   autofundXrp: true,
   autofundXrpAmount: 50,
   autofundAccountPin: '@CsmRxaR9UDXWfwy.rMv',
```
billguo99, line 43: what's the plan for the xrpIssuer account pin for Palau Prod?

billguo99 (approved):

PR #485, opened 2022-12-14
fix(web-client): Change auth_map type to Object type.

IscoRuta98:
### Description

- With the current implementation, the user's answers to  security questions are not saved correctly as the JavaScript message pack library does not seem to support ES6 Maps. More details of this issue can be found on this [link](https://github.com/msgpack/msgpack-javascript/issues/138).
- To fix this issue, the auth_map's type is converted (from a Map<string, string> type) to a Object type.


billguo99:
Have you tested this change locally to check that this change works?

PR #493, opened 2022-12-19
fix(web-client): display wait message while creating wallet.

sonasi:
Displays a busy icon while creating a wallet (if autofund is enabled). Also catch any errors that occur when submitting the autofund transaction and display a message "There was a problem creating the wallet, please try again.", then forward the user to the home page rather than showing it as a success and showing the defunct wallet address.


billguo99:
Rebase main and add RouterTestingModule to the imports for spec files that require it.

web-client/src/app/views/register/register.page.ts
```diff
@@ -108,22 +122,36 @@ export class RegisterPage implements OnDestroy {
...
+            { message: 'Creating Wallet' },
+            () =>
+              this.sessionXrplService.sendAutoFunds(
+                // wallet.xrpl_account.address_base58,
+                wallet_id,
+                autoFundAmount
+              )
           );
+          const amount = assetAmountXrp(autoFundAmount);
+          //const amount = convertFromLedgerToAssetAmountXrp(autoFundAmount);
```
billguo99, line 142: can remove this comment line

web-client/src/app/state/session-xrpl.service.ts
```diff
@@ -161,9 +162,10 @@ export class SessionXrplService {
     );
 
     const txResponse = await this.submitTransaction(txnSignedEncoded);
-    const txSucceeded = checkTxResponseSucceeded(txResponse);
+    await this.loadAccountData();
```
billguo99, line 165: Do we need to call `loadAccounData()`?

billguo99: This might be causing the Auto Logout problems.

billguo99: Nvm, this is not the cause of the auto logout feature. But do we need to call `loadAcountData()`?

billguo99:
Auto logout checks `this.sessionQuery.isActiveSession()` to check if there's an "active user session". Create Wallet updates the session with `this.sessionStore.update({ wallet, pin });` after it successfully creates a wallet. That is why auto logout comes into place when user has successfully created a wallet.

billguo99:
it seems you need to add `RouterTestingModule` for `AssetAccordionComponent` as well.

web-client/src/app/state/session-xrpl.service.ts
```diff
@@ -21,13 +22,13 @@ import { environment } from 'src/environments/environment';
 import { ifDefined } from 'src/helpers/helpers';
 import { TransactionSigned, TransactionToSign } from 'src/schema/actions';
 import * as xrpl from 'xrpl';
+import { TxResponse } from 'xrpl';
```
jongbonga, line 25: to keep the imports consistent, can you remove this line and use `xrpl.TxResponse` where you need it?

web-client/src/app/state/session-xrpl.service.ts
```diff
@@ -139,7 +140,7 @@ export class SessionXrplService {
   async sendAutoFunds(
     receiverId: string,
     amount: number
-  ): Promise<xrpl.TxResponse> {
+  ): Promise<{ xrplResult: TxResponse }> {
```
jongbonga, line 143: this can be replaced with `xrpl.TxResponse`

web-client/src/app/views/register/register.page.ts
```diff
@@ -7,13 +7,25 @@ import {
...
+  formatAssetSymbol,
+} from 'src/app/utils/assets/assets.common';
+import { assetAmountXrp } from 'src/app/utils/assets/assets.xrp';
+import { withLoadingOverlayOpts } from 'src/app/utils/loading.helpers';
+import { SwalHelper } from 'src/app/utils/notification/swal-helper';
 import { environment } from 'src/environments/environment';
 import SwiperCore, { Pagination } from 'swiper';
 import { SwiperComponent } from 'swiper/angular';
+import * as xrpl from 'xrpl';
+import { TxResponse } from 'xrpl';
```
jongbonga, line 28: same here

web-client/src/app/views/register/register.page.ts
```diff
@@ -108,22 +122,35 @@ export class RegisterPage implements OnDestroy {
...
         }
+        this.router.navigate(['/print-wallet']);
       } catch (err) {
-        // TODO: error handling
-        console.log(err);
+        this.notification.swal.fire({
+          icon: 'error',
+          titleText: 'Wallet Not Created!',
+          text: 'There was a problem creating your wallet, please try again.',
+          html: '<div >There was a problem creating your wallet, please try again.</div>',
```
jongbonga, line 150: text is already used, there is no need for html

web-client/src/app/views/register/register.page.ts
```diff
@@ -149,4 +176,75 @@ export class RegisterPage implements OnDestroy {
...
+    txUrlPrefix?: string;
+  }): void {
+    const txIdHtml = txUrlPrefix
+      ? `<a href="${txUrlPrefix}${txId}" target="_blank" >${txId}</a>`
+      : `${txId}`;
+    this.notification.swal.fire({
+      icon: 'success',
+      titleText: 'Wallet Created!',
+      text: 'Your wallet was successfully created.',
+      html: `<div >
```
jongbonga, line 221: you need to choose 1, you can't have both text and html at the same time, [html will be used if both are provided](https://sweetalert2.github.io/#html)

jongbonga (changes requested):

web-client/src/app/views/register/register.page.ts
sonasi: Going with html here due to formatting.

IscoRuta98 (approved):

jongbonga (approved):

PR #499, opened 2023-01-06
fix(web-client): update prod feature flags

sonasi:
Update feature flags for auto-logout and pull payments.

### DIFF
--- a/web-client/src/environments/environment.prod.ts
+++ b/web-client/src/environments/environment.prod.ts
@@ -40,7 +40,7 @@ export const environment: Environment = {
   hideXrpBalance: false,
   autofundXrp: true,
   hidePullPayment: false,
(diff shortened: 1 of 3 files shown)


web-client/src/environments/environment.prod.ts
```diff
@@ -39,7 +39,7 @@ export const environment: Environment = {
   xrpIssuer: 'rGqPqDW7PDNtRPyjFQ7fR47ojc3nUA5uBY',
   hideXrpBalance: false,
   autofundXrp: true,
-  hidePullPayment: true,
+  hidePullPayment: false,
   autoLogout: false,
```
jdvlio, line 43: Should auto logout still be disabled here?

jdvlio (changes requested):

PR #500, opened 2023-01-12
Quick access Functionality.

IscoRuta98:
### Quick access Functionality.
[Click up Task](https://app.clickup.com/t/31pm07a)


sonasi (approved):
If you have a saved address, the interface is a bit confusing. It would be better if, when you clicked anywhere in the box of the saved address, if it just filled in the form with the address. Finding the copy icon wasn't intuitive for me. 

Otherwise everything works good.

PR #511, opened 2023-01-27
refactor(web-client): show XRP transaction history on staging & demo environments.

IscoRuta98:
ClickupTask: https://app.clickup.com/t/31pkxve


sonasi (approved):

jongbonga (commented):
I think updating centrally like what I suggested is a better way of implementing this

PR #515, opened 2023-02-07
chore (web-client): remove kyc from palau env

sonasi:


IscoRuta98 (commented):
Looks good! KYC is hidden on Palau's staging and demo environments. 
Just need to run `npm run format` for the web-client test (specifically `prettier`) check to pass.

IscoRuta98 (approved):

PR #532, opened 2023-02-23
Autofund api function

sonasi:
Add endpoint to call autofund API.


jdvlio (approved):
Seems good as far as I can tell :+1:

PR #537, opened 2023-02-24
Autofund API Error handling

sonasi:
Catch ripple transaction errors and return the proper CORS header variables. This gives a 503 "Autofund temporarily unavailable" error instead of a 500 if anything goes wrong with the transaction.  This status code's retry functionality could later be leveraged in a future PR.

Remove unnecessary environment.ts variables for the old autofund method.


backend-services/src/data_service/operations/autofund.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+
+        my_tx_payment_signed = await safe_sign_and_autofill_transaction(
+            my_tx_payment, sender_wallet, client
+        )
+        tx_response = await send_reliable_submission(my_tx_payment_signed, client)
+    except:
+        raise HTTPException(
+            status_code=404, detail="Error submitting autofund transaction."
+        )
```
jdvlio, line 42: Due to the transient nature of the exceptions raised when submitting transactions we might want to use a [503 status code][503] instead.

[503]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503

```suggestion
        raise HTTPException(
            status_code=503, detail="Autofund transactions temporarily unavailable."
        )
```

backend-services/src/data_service/operations/autofund.py
```diff
@@ -0,0 +1,44 @@
...
+
+    return {"Access-Control-Allow-Origin": app_settings.primary_origin}
```
jdvlio, line 44: By raising the HTTPException, FastAPI will now know to submit the CORS headers so no value need be returned anymore.  The return type for this function can be `None` just as it was before.

backend-services/src/data_service/operations/autofund.py
```diff
@@ -0,0 +1,44 @@
...
+        my_tx_payment = Payment(
+            account=sender_account,
+            amount=xrp_to_drops(app_settings.autofund_amount),
+            destination=wallet_id,
+        )
+
+        my_tx_payment_signed = await safe_sign_and_autofill_transaction(
+            my_tx_payment, sender_wallet, client
+        )
+        tx_response = await send_reliable_submission(my_tx_payment_signed, client)
```
jdvlio, line 38: I'm uncertain whether it's necessary to assign the response on the last line to a variable or not.  If not then we should consider simplifying it.

Only transaction submission can raise an exception here so this can be reduced to the following.

```suggestion
        app_settings = AppSettings()
        client = JsonRpcClient(app_settings.autofund_server)

        sender_wallet = Wallet(
            app_settings.autofund_key, app_settings.autofund_sequence
        )
        sender_account = sender_wallet.classic_address

        my_tx_payment = Payment(
            account=sender_account,
            amount=xrp_to_drops(app_settings.autofund_amount),
            destination=wallet_id,
        )

        my_tx_payment_signed = await safe_sign_and_autofill_transaction(
            my_tx_payment, sender_wallet, client
        )
    try:
        tx_response = await send_reliable_submission(my_tx_payment_signed, client)
```

backend-services/src/web_asgi/settings.py
```diff
@@ -12,3 +12,8 @@ class AppSettings(Settings):
     staging_mode: bool = False
     wallet_db_name: str
     wallet_db_connection_string: MongoDsn
+    autofund_account: str
```
jdvlio, line 15: We could make use of `pydantic` validation here to make life easier for ourselves.

```suggestion
    autofund_account: HttpUrl
```

jdvlio (changes requested):

backend-services/src/data_service/operations/autofund.py
sonasi: Yep, you're right on both counts.

backend-services/src/web_asgi/settings.py
sonasi: you mean for autofund_server right, not autofund_account?

backend-services/src/data_service/operations/autofund.py
```diff
@@ -0,0 +1,44 @@
...
+            destination=wallet_id,
+        )
+
+    try:
+        client = JsonRpcClient(app_settings.autofund_server)
+        my_tx_payment_signed = await safe_sign_and_autofill_transaction(
+            my_tx_payment, sender_wallet, client
+        )
+        tx_response = await send_reliable_submission(my_tx_payment_signed, client)
+    except:
```
jdvlio, line 39: This should satisfy the CI as far as I'm aware, though a standard library import may also be needed.  The `ruff` documentation explains why it points out "bare" exceptions".

```suggestion
    except Exception:
```

jdvlio (changes requested):

backend-services/src/web_asgi/settings.py
jdvlio: apologies yes

backend-services/src/web_asgi/settings.py
```diff
@@ -12,3 +12,8 @@ class AppSettings(Settings):
...
+    autofund_key: str
+    autofund_server: str
```
jdvlio, line 17: For real this time.  Well done catching my error!
```suggestion
    autofund_server: HttpUrl
```

jdvlio (changes requested):

jdvlio (approved):

PR #543, opened 2023-03-08
Submit bug fix

IscoRuta98:
ClickUp Task: https://app.clickup.com/t/862jdc58n


sonasi (approved):
I reproduced the double click problem without this PR and then tested the changes here. At least for iOS the button is locked out after the first click.

PR #544, opened 2023-03-27
fix(web-client): Error Message on Wrong address

IscoRuta98:
### Error Message on Wrong address and/or pin entered

Users who mistype their wallet address into the Open Wallet function are still asked for a pin, and the wallet attempts to open but simply comes up with an error “Unexpected Error Occurred” …. they think this is a technical failure as there is no indication the information the entered was incorrect. We should have better capture and communication to the user that the address input was invalid.

### Testing the bug fix
1. Checkout branch `login-error-message`.
2. Launch the Nautilus Wallet App locally.
3. Login with an invalid address, and the below error message should appear.

image


jdvlio (approved):
:+1:

PR #545, opened 2023-03-28
Fix remember address issue & a confirmation dialog box before removing a saved address 

IscoRuta98:
### Remember Address Issue

When user clicks on ""remember address"", once pin entered, the wallet requests user to enter a nickname and press Ok. If users changes his mind and hits cancel, the wallet is still saved with nickname ""undefined"". In this case, it not only creates an Unknown item in the saved addresses list but duplicates the last entry in the existing saved addresses list where ones already exist. Lastly it allows users to save addresses they already have in their saved addresses which causes confusion and long duplicate lists in the saved addresses.

### Remove saved address needs a confirmation dialog box

Current functionality for the removal of saved addresses is to only click the red bin icon on the entry - and it is instantly removed. This can be done far too easily by mistake on a mobile phone, either as the user navigates the screen or tries to copy the saved address (as this requires pressing a button next to delete). Users will heavily depend on their wallet address being in the saved addresses and raise issues when this gets deleted. The functionality should require confirmation through a yes/no dialog box before deleting the entry.

### DIFF
--- a/web-client/src/app/components/pin-entry/pin-entry.component.ts
+++ b/web-client/src/app/components/pin-entry/pin-entry.component.ts
@@ -110,33 +110,35 @@ export class PinEntryComponent implements OnInit {
       this.walletAccessPage.address !== undefined
         ? this.walletAccessPage.address
         : '';
-    console.log(saveWalletAddress);
     try {
-      await this.notification.swal
-        .fire({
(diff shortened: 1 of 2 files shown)


jdvlio (approved):
Seems perfect :+1:

PR #547, opened 2023-04-12
refactor(web-client): user logging in using quick access.

IscoRuta98:
### Background

The need to click a copy icon on saved addresses and then paste the address into the "Enter wallet address" text box above is very fiddly on mobile phones. It would be much more intuitive and easy for users if they could click on the saved address and it either automatically populates the text field, or even better progresses to the Pin entry screen based on that address.

### Acceptance criteria

The user can click on the saved address and be taken to the pin entry directly without needing to enter the saved wallet address


jdvlio (approved):
Seems good.  You might want to change the "icon" though to avoid confusion.

PR #548, opened 2023-04-14
feat(backend-services): initial invite codes

jdvlio:
Initial implementation of invite code functionality.  Invite codes are used to act as a gate for new users for the duration of the pilots.


IscoRuta98 (approved):
Everything looks good to me.

PR #549, opened 2023-04-18
fix(web-client): show local time on history txs.

IscoRuta98:
### User Story

As a wallet holder I want to be able to see my transactions in Local Time on transaction History so that I do not have to convert the time myself

### Background

The time shown on transaction history is currently in UTC. These times need to match the time zone the user is currently in.

### Acceptance Criteria:

- The time displayed changes to reflect the time in the time zone the user is currently in


web-client/src/app/pipes/xrp-date.pipe.ts
```diff
@@ -1,14 +1,15 @@
...
 })
 export class XrpDatePipe implements PipeTransform {
   transform(date: number | undefined): string {
     if (date) {
-      const from = new Date('01/01/2000');
-      const finalDate = new Date(from.setSeconds(from.getSeconds() + date));
-      return new Date(finalDate).toLocaleString();
+      const from = moment.utc('01/01/2000', 'DD/MM/YYYY');
+      const finalDate = moment(from).add(date, 'seconds');
+      return finalDate.local().format('YYYY-MM-DD HH:mm:ss');
```
jongbonga, line 12: Are you not supposed to do `local(true)`? 

as per the [doc](https://momentjs.com/docs/#/manipulating/local/)?

jdvlio (approved):
Seems good :+1:

PR #555, opened 2023-04-28
Fix frontend prompt user to save nickname

IscoRuta98:
Aim of this PR is to fix the following [issue](https://ntls.atlassian.net/browse/NW-113?focusedCommentId=10020): Make sure a user cannot save an empty wallet nickname. They can chose to cancel, and this won't cause duplication in the quick access component.
image

### DIFF
--- a/web-client/src/app/state/qAccess/q-access.service.ts
+++ b/web-client/src/app/state/qAccess/q-access.service.ts
@@ -131,7 +131,10 @@ export class QAccessService implements OnDestroy {
         });
       }
     } catch (error) {
-      console.log(error);
+      await this.notification.swal.fire({
+        icon: 'error',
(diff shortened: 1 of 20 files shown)


web-client/src/app/state/qAccess/q-access.service.ts
```diff
@@ -124,7 +131,10 @@ export class QAccessService implements OnDestroy {
         });
       }
     } catch (error) {
-      console.log(error);
+      await this.notification.swal.fire({
+        icon: 'error',
+        text: 'An unexpected error occured when saving your wallet address.',
```
jongbonga, line 136: does this error prevent the user from accessing their account? if yes, maybe add a footer saying: please try again

PR #557, opened 2023-05-02
refactor(web-client): nickname prompt before login

IscoRuta98:
### Background

The enter Wallet Nickname dialog box appears on clicking Confirm on the PIN screen, the user then clicks on the text box and starts entering a name and as the app progresses in the background the text box loses focus so the user typing does not work. Note that this is very awkward in Palau where the lag due to internet is much longer (it all happens over many seconds as opposed to 1-2 seconds).

### Acceptance Criteria

Wait for the wallet nickname to be entered before opening the app

Prevent the same wallet from being saved twice. If the wallet address already exists in local storage do not display the save address checkbox to the user again

### DIFF
--- a/web-client/src/app/state/qAccess/q-access.service.ts
+++ b/web-client/src/app/state/qAccess/q-access.service.ts
@@ -102,36 +102,44 @@ export class QAccessService implements OnDestroy {
     });
   }
 
-  async saveQuickAccess(address: string | undefined) {
+  async saveQuickAccess(
+    address: string | undefined,
+    promptForNickname: boolean = true
+  ) {
     const saveWalletAddress: string = address !== undefined ? address : '';
     try {
-      const result = await this.notification.swal.fire({
-        titleText: 'Enter Wallet Nickname.',
-        input: 'text',
-        inputAttributes: {
-          autocapitalize: 'off',
-          autocorrect: 'off',
-        },
-        focusConfirm: false,
-        confirmButtonText: 'Save Wallet Address',
-        showCancelButton: true,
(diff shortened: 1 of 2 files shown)


web-client/src/app/state/qAccess/q-access.service.ts
```diff
@@ -102,36 +102,44 @@ export class QAccessService implements OnDestroy {
     });
   }
 
-  async saveQuickAccess(address: string | undefined) {
+  async saveQuickAccess(
+    address: string | undefined,
+    promptForNickname: boolean = true
```
jongbonga, line 107: why is the default for `promptForNickname` true?

IscoRuta98: This is to ensure that the prompt does remain open in the case the user tries to submit an empty name , and are faced with an error.

jongbonga: is there use cases where it can be false?

IscoRuta98: No, there currently is not a case where it is false.

IscoRuta98: I thought I'd do something similar for the invite code prompt, and cater for potential use cases. For now, there isn't a use case to have the parameter false. Do you advise that I remove it?

jongbonga: no it's okay

jongbonga (approved):

PR #562, opened 2023-05-12
feat(backend-services): initial 2FA implementation

jdvlio:
Initial 2FA implementation using Twilio Verify


backend-services/src/data_service/schema/entities.py
```diff
@@ -35,3 +35,17 @@ class Invite(Model):
 
     code: str
     redeemed: bool = False
+
+
+class OtpLimit(Model):
```
IscoRuta98, line 40: I think there should be two models because there are two different triggers for OTP:
1. If the amount sent is greater than the threshold limit.
2. If a wallet address has been saved for 2FA, regardless of the limit, 2FA should be triggered.

backend-services/src/web_asgi/main.py
```diff
@@ -6,15 +6,36 @@
...
+from data_service.operations.otp import otp_limits
+from data_service.schema.actions import (
+    CreateBookmark,
+    CreateOtpLimit,
+    DeleteBookmark,
+    DeleteOtpLimit,
+    RedeemInvite,
+)
+from data_service.schema.entities import Bookmark, BookmarkList, Invite, OtpLimitList
+from src.auth_service.operations.twilio_2fa import (
```
IscoRuta98, line 35: The twilio_2fa python file (i.e. src/auth_service/operations/twilio_2fa.py) seems to be missing from this PR.

IscoRuta98 (commented):
I think the PR is on the right track, except for two areas as highlighted in the below comments.

backend-services/src/web_asgi/main.py
```diff
@@ -59,6 +92,31 @@ async def post_bookmark_create(request: CreateBookmark) -> Bookmark:
     return await create_bookmark(mongo_engine, request)
 
 
+@app.post("/otp/limit/create", response_model=None, status_code=status.HTTP_201_CREATED)
```
IscoRuta98, line 95: I think there should be a PUT request for users who want to update their limit's threshold.

web-server/sgx-wallet-impl/src/schema/entities.rs
```diff
@@ -90,6 +92,7 @@ pub struct WalletStorable {
 
     pub owner_name: String,
     pub phone_number: Option<String>,
+    pub otp_phone_number: Option<String>,
```
IscoRuta98, line 95: When user creates a wallet for the first time, the OTP phone number should default to the user's phone number.

PR #569, opened 2023-05-24
feat(web-client): Initial 2FA Frontend

IscoRuta98:
### Background:

As a wallet owner I want to ensure that I can securely transact.

### Acceptance Criteria: 

- User can set up 2FA for transactions with the following options
- Set up cell number for 2FA
- Set up limits per account for which amounts equal to or exceeding will require 2FA.
- Set up limits for payments to specific wallet addresses which will require 2FA. 
- On making a payment which triggers one of the rules sets as per the above, the wallet owner will receive an OTP and be prompted to enter this in in order to conclude the payment. (pop up)
- On changing limits or removing a wallet address the user should be prompted to complete 2FA to complete the action

### DIFF
--- a/web-client/src/app/services/otp-prompt.service.ts
+++ b/web-client/src/app/services/otp-prompt.service.ts
@@ -1,8 +1,8 @@
 import { Injectable } from '@angular/core';
-import { SwalHelper } from '../utils/notification/swal-helper';
 import { CapacitorHttp } from '@capacitor/core';
-import { SessionQuery } from '../state/session.query';
 import { createUrlWith } from 'src/app/utils/http.helpers';
+import { SessionQuery } from '../state/session.query';
+import { SwalHelper } from '../utils/notification/swal-helper';
 
 const headers = {
   'Content-Type': 'application/json',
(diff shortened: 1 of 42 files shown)


web-client/src/app/services/otp-prompt.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    this.sendOtp(phoneNumberToSend);
+    const lastFourDigits = phoneNumberToSend?.slice(-4);
+    const maskedPhoneNumber = `*** *** ${lastFourDigits}`;
+
+    const { value: password } = await this.notification.swal.fire({
+      titleText: 'Enter the OTP',
+      text:
+        'A one time pin has been sent to the number ending with ' +
+        maskedPhoneNumber +
+        '. Please enter the 4 digit code below',
```
billguo99, line 37: I think it's a 6 digit code. :)

web-client/src/app/state/otpLimits/otp-limits.service.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    if (wallet_id) {
+      return await CapacitorHttp.get({
+        headers,
+        url: createUrlWith('otp/limit/triggers'),
+        params: { wallet_id },
+      })
+        .then(({ status, data }) => {
+          if (status === 200) {
+            console.log(data);
```
billguo99, line 36: Should we remove this? Not a big deal though.

billguo99:
Yeah there's still something buggy about the wallet addresses. When I add two wallet addresses in a row, it doesn't show two has been added.

billguo99 (approved):

PR #581, opened 2023-05-31
fix(web-client): check wallet id on scan success

jongbonga:


billguo99:
rebase with main :)

billguo99:
Where do you access the scan page to scan a QR code? I clicked "open wallet" but it doesn't give option to scan QR code.

jongbonga:
This PR addresses something else. If you are running the code locally, you can remove `hasCamera` on the wallet access template to test

billguo99:
> This PR addresses something else. If you are running the code locally, you can remove `hasCamera` on the wallet access template to test

What does this PR address?
Is this only for mobile app?
Which file should I remove `hasCamera`?

billguo99:
I removed `hasCamera` ng-container block and Access wallet with QR code option worked. I don't have a wallet address QR code to test so that's as far as I've tested. The option comes up and the camera webcam works.

billguo99 (approved):

IscoRuta98 (approved):

PR #582, opened 2023-06-01
refactor(web-client): check if device has camera

jongbonga:
This update will correctly show or hide the qr code access button


web-client/src/app/utils/camara.helpers.ts
```diff
@@ -0,0 +1,12 @@
+import { Capacitor } from '@capacitor/core';
+
+export const deviceHasCamera = async () => {
+  if (Capacitor.isNativePlatform()) {
+    return Capacitor.isPluginAvailable('Camera');
+  } else {
+    const devices = await navigator?.mediaDevices.enumerateDevices();
+    console.log(devices);
```
IscoRuta98, line 8: Is this `console.log()` part of debugging?
If so, perhaps it should be removed?

IscoRuta98 (approved):

PR #583, opened 2023-06-02
fix(web-client): Fix 2FA bugs

IscoRuta98:
The aim of this PR is to address the following bugs:
 1. Wallet address input  for OTP recipients is accepting any text input. It should only accept valid addresses.
 2. Update OTP cell number does not work.
 3. Current limits aren't showing. User should be able to see current limit for a given account.
 4. 2FA needs to fire when changing any triggers.


web-client/src/app/views/triggers/triggers.page.ts
```diff
@@ -70,25 +85,100 @@ export class TriggersPage implements OnInit {
...
+          this.notification.showUnexpectedFailureWarning();
+        }
+      }
+    );
+  }
+
+  /** Validated {@link address}, or `undefined`. */
+  validatedAddress(address: string): string {
+    const trimmed = address?.trim();
+    return trimmed === '' ? '' : trimmed;
```
jongbonga, line 126: Can you please explain what this does?

IscoRuta98: This function just makes sure the user does not submit an empty wallet address. In the event it does, an 'Invalid Address' popup message appears.

web-client/src/app/state/session.service.ts
```diff
@@ -144,6 +144,15 @@ export class SessionService {
     };
     const result: UpdateOtpPhoneNumberResult =
       await this.enclaveService.updateOtpPhoneNumber(request);
+
+    if ('Updated' in result) {
+      this.sessionStore.update((state) => {
+        if (state.wallet) {
```
jongbonga, line 150: no need to do this because you can only update the session when you are logged and you can't have a session without a wallet

web-client/src/app/views/triggers/triggers.page.html
```diff
@@ -47,7 +50,7 @@ <h1>Limits</h1>
       <ng-template>
         <ion-list lines="full">
           <ng-container *ngFor="let option of paymentOptions">
-            <ion-item button (click)="selectedOption = option" class="">
+            <ion-item button (click)="selectAccount(option)" class="">
```
jongbonga, line 53: `class=""`?

web-client/src/app/views/triggers/triggers.page.ts
jongbonga: But why are you returning `''` ?

web-client/src/app/views/two-factor-authentication/two-factor-authentication.page.ts
```diff
@@ -69,6 +76,12 @@ export class TwoFactorAuthenticationPage implements OnInit {
     }
   }
 
+  ngOnDestroy() {
```
jongbonga, line 79: You don't have to use `ngOnDestroy`, we have `UntilDestroyed` utility for unsubscribe automatically for us

jongbonga (changes requested):
Please see the comments I left

web-client/src/app/views/triggers/triggers.page.ts
IscoRuta98: Because of the `xrpl.isValidAddress()` and `algosdk.isValidAddress()` functions from the respective xrpl and algo SDKs. They only take strings variables when checking the validity of a wallet address. Instead of returning undefined, I am returning an empty (`''`) string, so that I can use the above functions.

jongbonga: But `isValidAddress()` will always be `false` if you pass an empty string, why bother passing it?

web-client/src/app/views/triggers/triggers.page.html
IscoRuta98: Removed.

web-client/src/app/views/triggers/triggers.page.ts
jongbonga: we should check if the address exists, validating an empty string is a pointless exercise, it's like doing `if(false)`

IscoRuta98: Resolved. I've removed the checker that checks for an empty string.

web-client/src/app/views/two-factor-authentication/two-factor-authentication.page.ts
```diff
@@ -56,7 +63,11 @@ export class TwoFactorAuthenticationPage implements OnInit {
     });
   }
 
-  ngOnInit() {}
+  ngOnInit() {
+    this.sessionQuery.wallet.pipe(untilDestroyed(this)).subscribe((wallet) => {
```
jongbonga, line 67: why don't you just subscribe to this on the constructor and pull the wallet id from it instead of subscribing on both `wallet` & `walletId`?

web-client/src/app/views/triggers/triggers.page.html
```diff
@@ -28,6 +28,9 @@ <h4 class="font-bold">Set Triggers</h4>
     <ion-list-header>
       <ion-label color="primary" class="ion-padding-horizontal ion-text-center">
         <h1>Limits</h1>
+        <ion-text color="light"
+          >Please choose an account to set a Limit for.</ion-text
```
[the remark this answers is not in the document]
jongbonga: No

Just format it to be `</ion-text>` instead of `</ion-text
>`

web-client/src/app/views/triggers/triggers.page.ts
```diff
@@ -70,25 +85,85 @@ export class TriggersPage implements OnInit {
...
     if (form.valid) {
       const otpRecipient = form.value.otpRecipient;
-      await this.otpRecipientsService
-        .createOtpRecipient(otpRecipient)
-        .then((success) => {
-          if (success) {
-            form.reset();
+      if (addressType(otpRecipient)) {
+        const otpAttempt = await this.otpPromptService.requestOTP();
+        if (!otpAttempt) {
```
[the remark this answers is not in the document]
jongbonga: Think or certain?

web-client/src/app/views/two-factor-authentication/two-factor-authentication.page.ts
```diff
@@ -18,6 +21,8 @@ export class TwoFactorAuthenticationPage implements OnInit {
   isOpening = false;
   registrationForm: FormGroup;
   walletId: string | undefined;
+  otpPhoneNumber: string | undefined;
+  subscription: Subscription | undefined;
```
jongbonga, line 25: are you still using this?

jongbonga (approved):

web-client/src/app/views/two-factor-authentication/two-factor-authentication.page.ts
IscoRuta98: No we are not using subscription variable.
I've removed it.

PR #588, opened 2023-06-09
fix(web-client): pull bookmarks when wallet opens

jongbonga:
This addresses the issue where the tx history will not display the name because there bookmark list isn't available yet

### DIFF
--- a/web-client/src/app/views/wallet/wallet.page.ts
+++ b/web-client/src/app/views/wallet/wallet.page.ts
@@ -1,6 +1,7 @@
 import { Component, Input, OnInit } from '@angular/core';
 import { UntilDestroy } from '@ngneat/until-destroy';
 import { combineLatest, map, Observable } from 'rxjs';
+import { BookmarkService } from 'src/app/state/bookmark';
 import { OtpLimitsService } from 'src/app/state/otpLimits';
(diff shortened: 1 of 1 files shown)


web-client/src/app/views/wallet/wallet.page.ts
```diff
@@ -95,6 +97,7 @@ export class WalletPage implements OnInit {
   }
 
   async ionViewWillEnter() {
+    this.bookmarkService.getBookmarks();
```
IscoRuta98, line 100: Given this is an `async` function, shouldn't the `await` keyword be  used before calling the function.

`await this.bookmarkService.getBookmarks();`

jongbonga: You use `await` if you need to use the results right away, but we don't for this case. We call  it like that so that the result are available in the store when we open the transaction history. 

If we needed the results on the wallet page, yes, we should probably use await, but we don't.

IscoRuta98: I see, thanks for the clarification. The PR looks good to go, the fix works!

IscoRuta98 (approved):

PR #589, opened 2023-06-12
fix(web-server): Fix backend Pin Reset accepting any security answers.

IscoRuta98:
This PR aims to fix to the following issue: [Forgot Pin | Accepting incorrect security questions](https://ntls.atlassian.net/jira/software/projects/NW/boards/1/backlog?selectedIssue=NW-198).

### DIFF
--- a/web-server/sgx-wallet-impl/src/schema/auth.rs
+++ b/web-server/sgx-wallet-impl/src/schema/auth.rs
@@ -46,6 +46,9 @@ impl WalletChallenge {
             Some(mut wallet_storable) => {
                 let stored_responses =
                     WalletAuthMap::from_iter(wallet_storable.auth_map.drain().into_iter());
+                if stored_responses.values().any(|response| response.is_empty()){
+                    return Err(AuthError::InvalidAttempt)
+                }
                 Ok(Self {
                     auth_attempts,
                     stored_responses,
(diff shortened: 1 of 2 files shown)


billguo99:
How did you test this fix?

IscoRuta98:
> How did you test this fix?

I have not tested it yet. I'd like to merge this to staging.ntls.io, and test through the frontend.

billguo99:
Spelling error in git commit message. "security"

web-server/sgx-wallet-impl/src/schema/auth.rs
```diff
@@ -46,6 +46,9 @@ impl WalletChallenge {
             Some(mut wallet_storable) => {
                 let stored_responses =
                     WalletAuthMap::from_iter(wallet_storable.auth_map.drain().into_iter());
+                if stored_responses.values().any(|response| response.is_empty()){
```
billguo99, line 49: This if-block is underneath the "new" function, so it'll only trigger when "new" WalletAuthMaps are generated.

Should your fix not be part of the "authenticate" function beneath?

I thought the error was that, when we authenticate the user's answers, if the user's answers were empty, it's still a success.
https://github.com/ntls-io/nautilus-wallet/blob/879f211cde83cd5d9d6ddffc54cea187ed93bde8/web-server/sgx-wallet-impl/src/wallet_operations/pin_reset.rs#L22

IscoRuta98: I do not think so. From my understanding the `new` function in the `WalletChallenge` struct essentially initializes an `auth` attempt by loading the stored responses from an existing wallet. Therefore. regardless of a new or old wallet, the process is the same.

billguo99 (approved):

PR #590, opened 2023-06-13
fix(web-client): Prevent auto-logout

IscoRuta98:
This PR aims to fix the following bug: [Prevent auto-logout when an error occurs](https://ntls.atlassian.net/jira/software/projects/NW/boards/1?selectedIssue=NW-178).

### DIFF
--- a/web-client/src/app/state/session-xrpl.service.ts
+++ b/web-client/src/app/state/session-xrpl.service.ts
@@ -552,7 +552,7 @@ export class SessionXrplService {
           })
           .then(({ isConfirmed }) => {
             if (isConfirmed) {
-              this.navCtrl.navigateRoot('/');
+              this.navCtrl.back();
             }
           });
         throw new Error('Local error:');
--- a/web-client/src/app/utils/errors/global-error-handler.ts
+++ b/web-client/src/app/utils/errors/global-error-handler.ts
@@ -35,7 +35,7 @@ export class GlobalErrorHandler implements ErrorHandler {
           })
           .then(() => {
             // XXX: See also modal closing code in LandingPage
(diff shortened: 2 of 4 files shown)


jongbonga (commented):
Do you know that `back()` just navigate one page back right?

Why do you navigate back on global error?

IscoRuta98:
> Do you know that `back()` just navigate one page back right?
> 
> Why do you navigate back on global error?

Yes, I do. The [Acceptance Criteria](https://ntls.atlassian.net/jira/software/projects/NW/boards/1?selectedIssue=NW-178) of the task states that the user should be returned back to the previous page.

web-client/src/app/state/session-xrpl.service.ts
```diff
@@ -552,7 +552,7 @@ export class SessionXrplService {
           })
           .then(({ isConfirmed }) => {
             if (isConfirmed) {
-              this.navCtrl.navigateRoot('/');
+              this.navCtrl.back();
```
jongbonga, line 555: This shouldn't be done in the service, it should be handled on the page. that way you have different behaviours instead of this one back navigation

web-client/src/app/utils/errors/global-error-handler.ts
```diff
@@ -35,7 +35,7 @@ export class GlobalErrorHandler implements ErrorHandler {
           })
           .then(() => {
             // XXX: See also modal closing code in LandingPage
-            this.navCtrl.navigateRoot('');
+            this.navCtrl.back();
```
jongbonga, line 38: I don't think this is the best thing to do.

jongbonga (approved):

PR #592, opened 2023-06-14
fix(web-client): wallet address validator for bookmark.

IscoRuta98:
The aim of this PR is to resolve the following issue: [PAY: Unsupported Address Screen](https://ntls.atlassian.net/jira/software/projects/NW/boards/1?selectedIssue=NW-205).

A wallet address is first checked for its validity before saving a new bookmark. That way, any payments made via bookmarks page will consist of valid addresses.


billguo99 (approved):
Looks good!

PR #596, opened 2023-06-22
fix(web-server): check for empty store in auth method.

IscoRuta98:
Fix for the following issue: [Forgot Pin | Accepting incorrect security questions](https://ntls.atlassian.net/jira/software/projects/NW/boards/1?selectedIssue=NW-198)

### DIFF
--- a/web-server/sgx-wallet-impl/src/schema/auth.rs
+++ b/web-server/sgx-wallet-impl/src/schema/auth.rs
@@ -55,10 +55,11 @@ impl WalletChallenge {
         }
     }
     pub fn authenticate(mut self) -> Result<ChallengeOutcome, AuthError> {
+        let stored_responses_copy = self.stored_responses.clone();
         let mut stored = self.stored_responses.drain();
         let mut mismatch_count = 0;
 
-        if stored.is_empty() {
+        if stored_responses_copy.is_empty() {
             return Err(AuthError::InvalidAttempt);
         }
 
--- a/web-server/sgx-wallet-impl/src/schema/auth.rs
+++ b/web-server/sgx-wallet-impl/src/schema/auth.rs
@@ -55,13 +55,9 @@ impl WalletChallenge {
         }
     }
     pub fn authenticate(mut self) -> Result<ChallengeOutcome, AuthError> {
-        let stored_responses_copy = self.stored_responses.clone();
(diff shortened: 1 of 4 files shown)


web-server/sgx-wallet-impl/src/schema/auth.rs
```diff
@@ -55,9 +55,14 @@ impl WalletChallenge {
         }
     }
     pub fn authenticate(mut self) -> Result<ChallengeOutcome, AuthError> {
+        let stored_responses_copy = self.stored_responses.clone();
```
billguo99, line 58: Do we have to clone it?

Can't we just check it?
```
    /*
     *  Authentication fails if there are any mismatches or if there are no
     *  stored responses.
     */
    if mismatch_count > 0 || self.stored_responses.is_empty() {
        return Ok(ChallengeOutcome::Failure);
    }
    Ok(ChallengeOutcome::Success)
}

```

IscoRuta98: I initially tried what you suggested however, I get `borrowing error`. Therefore, an alternative is to make a copy of the stored responses, then check the cloned version of the stored responses if its empty.

billguo99 (approved):

web-server/sgx-wallet-impl/src/schema/auth.rs
billguo99: We could've also done the following:
```
pub fn authenticate(mut self) -> Result<ChallengeOutcome, AuthError> {
    if self.stored_responses.is_empty() {
        return Ok(ChallengeOutcome::Failure);
    }

    let mut stored = self.stored_responses.drain();
    let mut mismatch_count = 0;

    while let Some((key, stored_secret)) = stored.next() {
        if let Some(attempt) = self.auth_attempts.get(&key) {
            // Comparison logic remains unchanged
            // ...
        } else {
            return Err(AuthError::InvalidAttempt);
        }
    }

    if mismatch_count > 0 {
        return Ok(ChallengeOutcome::Failure);
    }

    Ok(ChallengeOutcome::Success)
}
```

To address this issue, we can modify the code by extracting the is_empty() check before starting the loop to iterate over self.stored_responses. This way, we avoid the simultaneous mutable and immutable borrows.

PR #598, opened 2023-07-02
feat(web-client): Recurring payments feature.

IscoRuta98:
Implementation of Frontend Recurring Payment feature.

### DIFF
--- a/backend-services/src/data_service/operations/recurringpayment.py
+++ b/backend-services/src/data_service/operations/recurringpayment.py
@@ -1,0 +1,117 @@
+from datetime import date
+
+from bson import ObjectId
+from fastapi import HTTPException
+
+from common.types import WalletAddress
+from data_service.schema.actions import (
+    CreateRecurringPayment,
+    DeleteRecurringPayment,
+    UpdateLastPaidDate,
+    UpdateRecurringPayment,
+)
+from data_service.schema.entities import RecurringPayment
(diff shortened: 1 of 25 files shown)


billguo99:
Nice, just a few minor feedback:
* Can you remove the `console.log` that's not in error blocks?
* When inputting the days for frequency, can you indicate somewhere that the value for the frequency is in days?

One more fix:
* I saw the date values that you're saving to mongodb is basically Unix timestamp int values.

You'll need to use the following typescript function to convert a Date object to their ordinal value:
```
const _DAYS_IN_MONTH: number[] = [-1, 31, 28, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31];

function isLeapYear(year: number): boolean {
    // year -> 1 if leap year, else 0.
    return year % 4 === 0 && (year % 100 !== 0 || year % 400 === 0);
}

function getDateToOrdinal(date: Date): number {
    const year = date.getFullYear();
    const month = date.getMonth() + 1;
    const day = date.getDate();

    return (
        (year - 1) * 365 +
        Math.floor((year - 1) / 4) -
        Math.floor((year - 1) / 100) +
        Math.floor((year - 1) / 400) +
        _DAYS_IN_MONTH.slice(1, month).reduce((acc, curr) => acc + curr, 0) +
        (month > 2 && isLeapYear(year) ? 1 : 0) +
        day
    );
}
```

PR #599, opened 2023-07-04
fix(backend-services): change date values to be calculated from unix timestamps for recurring payments

billguo99:
date.ToOrdinal() method in Python is not compatible or very difficult to know what the implementation of the function is.

So I opted to just use int value of unix timestamps instead.


IscoRuta98 (approved):
LGTM!

billguo99:
Actually it's fine, I found the implementation in the Python docs - https://github.com/python/cpython/blob/3cc5c4a073d63534e45fc08d7906a77180e8aaa3/Lib/datetime.py#L63

PR #602, opened 2023-07-07
fix-recurring-pay-frontend

IscoRuta98:
This PR aims to change the following for Recurring Payment feature:

-  Frequency should be an option between:
  Every Day
  Every Week
  Every Month
  Quarterly
  
- After successful schedule, return to the main recurring payments page.

- Date picker icon should be white.


billguo99 (approved):
Looks good!

PR #604, opened 2023-07-13
chore: change palau prod port to 443

billguo99:
Use non-standard port as some networks blocks this.


Astewart1510:
Flawless work @billguo99

Astewart1510 (approved):

PR #606, opened 2023-07-18
fix-recurring-pay

IscoRuta98:
This PR aims to fix the following tickets:
- [NW-217: UI: Recurring Payment - Set Schedule - Dropdown](https://ntls.atlassian.net/jira/software/projects/NW/boards/1?selectedIssue=NW-217)
- [NW-218: Back Button on Recurring Payments Screen (After Creating Recurring Payment)](https://ntls.atlassian.net/jira/software/projects/NW/boards/1?selectedIssue=NW-218)

### DIFF
--- a/web-client/src/app/components/new-schedule-pay/new-schedule-pay.component.html
+++ b/web-client/src/app/components/new-schedule-pay/new-schedule-pay.component.html
@@ -13,6 +13,7 @@
                 justify="space-between"
                 label-placement="floating"
                 label="Select Frequency"
+                formControlName="frequency"
               >
                 <ion-select-option value="1">Every Day</ion-select-option>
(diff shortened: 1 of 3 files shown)


billguo99:
Getting the following error when trying to add a recurring payment:
image
I think the frequency input is not a valid integer, think we might be giving it a string.

billguo99:
image
Actually it's not getting passed to the request.

IscoRuta98:
> Getting the following error when trying to add a recurring payment: image I think the frequency input is not a valid integer, think we might be giving it a string.

I fixed this error, I accidently removed the formControlName. I've added it back.

billguo99 (approved):

PR #609, opened 2023-07-27
feat(web-server): initial implementation for account lock

billguo99:
https://ntls.atlassian.net/browse/NW-203?atlOrigin=eyJpIjoiZDg2Y2U1YjY1N2FiNGQ1ZGFkMDYxYmYxNDFhZjMxMzUiLCJwIjoiaiJ9

New feature to lock account if there are 3 attempts to log into account with incorrect pin.

This PR is for the Nautilus Wallet backend web-server.

Implementation plan:
web-server:
* add account_attempts: u32 to WalletStorable
* when open_wallet is called, increase account_attempts  by 1 if open_wallet fails for incorrect pin
* when open_wallet succeeds, reset account_attempts to 0
* when create_wallet is called for new wallet initialize account_attempts to 0
* for existing accounts, if account_attempts is none, or not there, add it to WalletStorable when open_wallet is called
* when open_wallet is called and account_attempts is equal to 3 or greater (should not be possible), return Account Lock Error with message: “You have failed to enter the correct pin 3 times. Please reset your pin in order to access your account.“ even if the pin is correct
* when pin_reset is called and OTP is successful, set new pin and reset account_attempts to 0.

### DIFF
--- a/web-server/sgx-wallet-impl/src/wallet_operations/update_otp_number.rs
+++ b/web-server/sgx-wallet-impl/src/wallet_operations/update_otp_number.rs
@@ -23,6 +23,9 @@ pub fn update_otp_phone_number(request: &UpdateOtpPhoneNumber) -> UpdateOtpPhone
                 ))
             }
             UnlockWalletError::IoError(err) => UpdateOtpPhoneNumberResult::Failed(err.to_string()),
+            UnlockWalletError::AccountLocked => {
+                UpdateOtpPhoneNumberResult::Failed("Account is locked".to_string())
+            }
         },
     }
 }
--- a/web-server/sgx-wallet-impl/src/lib.rs
+++ b/web-server/sgx-wallet-impl/src/lib.rs
@@ -3,6 +3,7 @@
 
 #[macro_use]
 extern crate sgx_tstd as std;
+extern crate alloc;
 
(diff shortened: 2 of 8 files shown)


web-server/sgx-wallet-impl/src/wallet_operations/update_otp_number.rs
```diff
@@ -23,6 +23,9 @@ pub fn update_otp_phone_number(request: &UpdateOtpPhoneNumber) -> UpdateOtpPhone
                 ))
             }
             UnlockWalletError::IoError(err) => UpdateOtpPhoneNumberResult::Failed(err.to_string()),
+            UnlockWalletError::AccountLocked => {
```
IscoRuta98, line 26: Why is there error handling for account lock in `update_opt_number` when a user would not be able to login if the person has tried logging in more than 3 times and has not reset their pin?

billguo99: the function `update_otp_phone_number` in update_otp_number.rs uses the method `unlock_wallet` and then `mutates` the otp number. Upon unlocking wallet, the user may experience Account Lock.

IscoRuta98 (approved):

PR #611, opened 2023-07-31
fix(web-client): diasble autologout in send & pull pay

IscoRuta98:
Aim of this PR is to disable Auto Logout when user is confirming transactions (for pull and pay payments).

### DIFF
--- a/web-client/src/app/views/pay/pay.page.ts
+++ b/web-client/src/app/views/pay/pay.page.ts
@@ -5,6 +5,7 @@ import { LoadingController, NavController } from '@ionic/angular';
 import { Observable, pluck } from 'rxjs';
 import { Payment } from 'src/app/components/pay/pay.component';
 import { TransactionConfirmation } from 'src/app/services/algosdk.utils';
+import { AutoLogoutService } from 'src/app/services/auto-logout.service';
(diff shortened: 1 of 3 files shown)


billguo99:
Have you tested this locally?

web-client/src/app/views/pay/pay.page.ts
```diff
@@ -145,19 +148,22 @@ export class PayPage implements OnInit {
         }
```
billguo99, line 148: Do we need to add `this.autoLogoutService.enableAutoLogout()` in the catch (error) block above on line 144?

IscoRuta98: No, it is not needed, will delete it in the latest commit.

billguo99: Wait, I think it's needed? I meant another catch block.
I saw you added the enableAutoLogout in the catch block and then I saw another one where you didn't add it.

IscoRuta98: I don't think so as the one I deleted should not have been added in the catch block as it (enableAutologout() method) was already added on the try block.

billguo99 (approved):
Cool!

PR #614, opened 2023-09-14
refactor: change autofund_amount from int to float

billguo99:
Changing `autofund_amount` from `int` to `float`  for backend-services.


IscoRuta (approved):
LGTM :)

